PR #2, opened 2016-09-05
Update Client.java

zjsunzone:
In my pro , use HelloPrx throw  Exception！


externl:
The code in the repository is correct. I suspect you may have been using Ice 3.6 jars, in which case you should be using the 3.6 java demos.

PR #6, opened 2017-12-07
3.7

bluelins:


pepone:
Hi @bluelins 

We merge 3.7 onto master from time to time, don't need a pull request for that, You should probably just use 3.7 branch

PR #8, opened 2018-02-22
ICE-8401 - Remove Ice.Application from python

rysamurai:


python/Ice/async/Client.py
```diff
@@ -5,7 +5,7 @@
 #
 # **********************************************************************
 
-import sys, os, traceback, threading, Ice
+import signal, sys, os, traceback, threading, Ice
```
pepone: signal is not used in this file, why do we need the import?

rysamurai: Thanks, I will fix it.

PR #40, opened 2019-05-20
Add async demo for Swift

ReeceHumphreys:


swift/Ice/async/Server/main.swift
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    func sayHelloAsync(delay: Int32, current _: Current) -> Promise<Void> {
+        return Promise { seal in
+            if delay == 0 {
+                print("Hello World!")
+                seal.fulfill(())
+            } else {
+                serialQueue.sync {
+                    guard self.workQueue.isSuspended == false else {
```
pepone: We should move this check inside the closure block passed to addOperation

swift/Ice/async/Server/main.swift
```diff
@@ -0,0 +1,84 @@
...
+        if let adapter = current.adapter {
+            adapter.getCommunicator().shutdown()
+        }
+    }
+}
+
+func run() -> Int32 {
+    do {
+        var args = CommandLine.arguments
+        let workQueue = OperationQueue()
```
pepone: We should initialize the workQueue just before it is used and after the `"too many arguments"` error check

swift/Ice/async/Client/main.swift
```diff
@@ -0,0 +1,87 @@
...
+        repeat {
+            print("==> ", terminator: "")
+            guard let l = readLine(strippingNewline: true) else {
+                return 0
+            }
+            line = l
+            guard let choice = selection(rawValue: line) else {
+                print("unknown command `\(line)'")
+                menu()
+                return 0
```
pepone: an unknown command should not terminate the loop, just show the menu a read a new line

swift/Ice/async/Client/main.swift
```diff
@@ -0,0 +1,87 @@
...
+            communicator.destroy()
+        }
+
+        guard args.count == 1 else {
+            print("too many arguments")
+            return 1
+        }
+
+        guard let base = try communicator.propertyToProxy("Hello.Proxy"),
+            let twoway = try checkedCast(prx: base.ice_twoway().ice_secure(false), type: HelloPrx.self)
```
pepone: Don't need to do `.ice_twoway().ice_secure(false)` here, just use 
```
let hello = try checkedCast(prx: base, type: HelloPrx.self)
```

swift/Ice/async/Server/main.swift
```diff
@@ -0,0 +1,84 @@
...
+                }
+            }
+        }
+    }
+
+    func shutdown(current: Current) throws {
+        print("Shutting down...")
+        // Setting this property to true prevents the queue from starting any queued operations,
+        // but already executing operations continue to execute.
+        serialQueue.sync {
```
pepone: We can dispatch this in the `concurrentQueue` with a barrier and get rid of the `serialQueue`

swift/Ice/async/Client/main.swift
```diff
@@ -0,0 +1,87 @@
...
+
+func menu() {
+    print(
+        """
+        usage:
+            i: send immediate greeting
+            d: send delayed greeting
+            s: shutdown server
+            x: exit
+        ?: help
```
pepone: `?: help` should be aligned with the other options

swift/Ice/async/Server/main.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    func sayHelloAsync(delay: Int32, current _: Current) -> Promise<Void> {
+        return Promise { seal in
+            if delay == 0 {
+                print("Hello World!")
+                seal.fulfill(())
+            } else {
+                serialQueue.sync {
+                    self.workQueue.addOperation {
+                        guard self.workQueue.isSuspended == false else {
```
externl: If the `workQueue` is suspended then this operation is added it will never execute. Instead of using `workedQueue.isSuspended` we should just add a `var destroyed: Bool`  and check it instead. That way the operation queue will always execute it's queued items.

pepone: Not sure we need a second var here, seems simpler to check `isSuspended` before adding to the queue too, so if the queue is already suspended we reject the promise right away, and operations that are already in the queue will throw once they are executed.

externl: That's fine too. One question is whether or not `isSuspended` its self is thread safe. If we access it outside of the operation we'll probably have to use an async dispatch.

pepone: I think we can schedule addOperation asynchronously in the `concurrentQueue`, get rid of the `serialQueue` and set `self.workQueue.isSuspended` from the `concurrentQueue` using a barrier

externl: One issue now is that if you add a delayed operation then call shutdown, odds are that the operation queue will be running while destroyed waits. 

So if you do:
```
> d
> d
> d
> s
```

on the server you'll see

```
Shutting down...
Belated Hello World!
Belated Hello World!
Belated Hello World!
```

externl: I discussed this with @bernardnormier. It's probably simplest to remove the operation queue and then dispatch asynchronously to the serial queue.

PR #41, opened 2019-06-04
Added callback demo for Glacier2

ReeceHumphreys:


bernardnormier:
The README.md is missing.

My other comments apply to this Swift version and also other language mappings - may as well get Swift right first!

- We should add more comments, with proper spelling (e.g. Command prompt)
- The callback receiver object ID name should be a UUID, with a comment that the server doesn't need to know about it
- Add a comment for "fake" explaining what it's about (the Glacier2 router routes bi-dir calls to objects in the client only when these objects have the correct Glacier2-issued category)
- we should not create/fill context when doing so does nothing:
  - for twoway requests: _ovrd has an effect only for oneways and batch oneways, while _fwd = t is the default behavior
  - for oneway requests: _fwd is not useful, but _ovrd is
 - for batch oneway requests: _fwd is useful with the default config. I would also just set _fwd explicitly to "O", it's more readable.
- in config.glacier2, we should remove:
{code}
#
# These two lines instruct Glacier2 to forward contexts both for
# regular routing, as well as for callbacks (reverse routing).
#
Glacier2.Client.ForwardContext=1
Glacier2.Server.ForwardContext=1
{code}
This demo does not use the forwarded contexts, so it's more confusing than anything to show these properties.

swift/Glacier2/callback/Sources/Client/main.swift
```diff
@@ -0,0 +1,188 @@
...
+                break
+            } catch {
+                print("Cannot created session: \(error)")
+            }
+        }
+
+        guard let acmTimeout = try? router.getACMTimeout(),
+            let connection = router.ice_getCachedConnection() else {
+            return 1
+        }
```
pepone: I don't thin we should be using `try?` here or `guard` for `acmTimeout`
```
let acmTimeout = try router.getACMTimeout()
guard let connection = router.ice_getCachedConnection() else {
    fatalError()
    return 1
}
```

swift/Glacier2/callback/Sources/Server/main.swift
```diff
@@ -0,0 +1,52 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+import Foundation
+import Glacier2
+import Ice
+import PromiseKit
```
pepone: You can remove `PromieKit` import is not used here

swift/Glacier2/callback/Sources/Server/main.swift
```diff
@@ -0,0 +1,52 @@
...
+        defer {
+            communicator.destroy()
+        }
+        guard args.count == 1 else {
+            print("too many arguments\n")
+            return 1
+        }
+        let adapter = try communicator.createObjectAdapter("Callback.Server")
+        let cb = CallbackI()
+        try adapter.add(servant: cb, id: stringToIdentity("callback"))
```
pepone: I think is simpler to create `CallbackI`  inline don't need for the `cb` reference

```
try adapter.add(servant: CallbackI(), id: stringToIdentity("callback"))
```

swift/Glacier2/callback/Sources/Callback.ice
```diff
@@ -0,0 +1,19 @@
...
+{
+void callback();
+}
+
+interface Callback
+{
+void initiateCallback(CallbackReceiver* proxy);
+void shutdown();
+}
+}
```
pepone: This file is not correctly indented , see https://github.com/zeroc-ice/ice-demos/blob/3.7/csharp/Glacier2/callback/Callback.ice

swift/Glacier2/callback/config.client
```diff
@@ -0,0 +1,53 @@
...
+#
+# 0 = no security tracing
+# 1 = trace messages
+#
+#IceSSL.Trace.Security=1
+
+#
+# SSL Configuration
+#
+Ice.Plugin.IceSSL=IceSSL:createIceSSL
```
pepone: For Swift it should be just `Ice.Plugin.IceSSL=1`

swift/Glacier2/callback/Sources/Client/main.swift
```diff
@@ -125,25 +126,15 @@ func run() -> Int32 {
             if let option = Option(rawValue: line) {
                 switch option {
                 case .twoway:
-                    var context = Context()
-                    context["_fwd"] = option.rawValue
-                    if !override.isEmpty {
-                        context["_ovrd"] = override
-                    }
-                    try twoway.initiateCallback(twowayR, context: context)
+                    try twoway.initiateCallback(twowayR, context: nil)
```
pepone: we can omit `context: nil` here

swift/Glacier2/callback/Sources/Client/main.swift
```diff
@@ -125,25 +126,15 @@ func run() -> Int32 {
...
                     try oneway.initiateCallback(onewayR, context: context)
                 case .batchOneway:
                     var context = Context()
-                    context["_fwd"] = option.rawValue
-                    if !override.isEmpty {
-                        context["_ovrd"] = override
-                    }
+                    context["_fwd"] = "O"
+                    context["_ovrd"] = (!override.isEmpty) ? override : nil
                     try batchOneway.initiateCallback(onewayR, context: context)
```
pepone: what about
```
let context: Context = override.isEmpty ? ["_fwd": "O"] : ["_fwd": "O", "_ovrd": override]
```

swift/Glacier2/callback/Sources/Client/main.swift
ReeceHumphreys: Yes this works better. However don't need the `return 1` as it will never be executed if `fatalError()` is called.

swift/Glacier2/callback/Sources/Client/main.swift
```diff
@@ -0,0 +1,188 @@
...
+            } catch {
+                print("Cannot created session: \(error)")
+            }
+        }
+
+        guard let acmTimeout = try? router.getACMTimeout(),
+            let connection = router.ice_getCachedConnection() else {
+            return 1
+        }
+        connection.setACM(timeout: acmTimeout, close: nil, heartbeat: .HeartbeatAlways)
```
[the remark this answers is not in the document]
pepone: ah right, this only works for remote invocations, that are implemented in a protocol extension, let's keep it as is

swift/Glacier2/callback/Sources/Server/main.swift
```diff
@@ -5,7 +5,6 @@
 import Foundation
 import Glacier2
```
pepone: Glacier2 import is not required in this file

swift/Glacier2/callback/Sources/Server/main.swift
```diff
@@ -38,8 +35,7 @@ func run() -> Int32 {
             return 1
         }
         let adapter = try communicator.createObjectAdapter("Callback.Server")
-        let cb = CallbackI()
-        try adapter.add(servant: cb, id: stringToIdentity("callback"))
+        try adapter.add(servant: CallbackI(), id: stringToIdentity("callback"))
```
pepone: Need update to use the new `Disp` structs, 

`try adapter.add(servant: CallbackDisp(CallbackI()), id: stringToIdentity("callback"))`

swift/Glacier2/callback/Sources/Client/main.swift
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        let oneway = twoway.ice_oneway()
+        let batchOneway = twoway.ice_batchOneway()
+        let adapter = try communicator.createObjectAdapterWithRouter(name: "", rtr: router)
+        try adapter.activate()
+        try adapter.add(servant: CallbackReceiverI(), id: callbackReceiverIdent)
+
+        // Should never be called for the fake identity.
+        try adapter.add(servant: CallbackReceiverI(), id: callbackReceiverFakeIdent)
```
pepone: need to update this to use the new `Disp` structs

bernardnormier:
You should also remove the `throws` from all the operation implementations, since none of them can throw anything.

PR #46, opened 2019-06-18
Updated Auto-layout and added MessageKit

ReeceHumphreys:


swift/iOS/Chat/ChatLayoutController.swift
```diff
@@ -0,0 +1,120 @@
+//
+//  ChatLayoutController.swift
+//  IOSChat
+//
+//  Created by Reece Humphreys on 6/18/19.
+//
+
```
pepone: You should add ZeroC copyright headers to all new files

swift/iOS/Chat/ChatLayoutController.swift
```diff
@@ -0,0 +1,120 @@
...
+    func messageTopLabelAttributedText(for message: MessageType, at _: IndexPath) -> NSAttributedString? {
+        let name = message.sender.displayName
+        let font = UIFont.preferredFont(forTextStyle: .caption1)
+        return NSAttributedString(string: name,
+                                  attributes: [NSAttributedString.Key.font: font])
+    }
+
+    func messageBottomLabelAttributedText(for message: MessageType, at _: IndexPath) -> NSAttributedString? {
+        let formatter: DateFormatter = {
```
pepone: I think it is better if we crate the formatter in the initializer and reuse it rather than create a new one each time this method is called,

Also why you need a closure for this? instead of just doing 
```
let formatter = DateFormatter()
formatter.dateStyle = .medium
```

swift/Rakefile
```diff
@@ -108,7 +108,8 @@ def create_commandline_targets(project, demo, aggregate_target)
 
             target_add_carthage_framework(target, :osx, "Ice.framework")
             target_add_carthage_framework(target, :osx, "PromiseKit.framework")
-
+	    target_add_carthage_framework(target, :osx, "MessageKit.framework")
+	    target_add_carthage_framework(target, :osx, "InputBarAccessoryView.framework")
             extra_frameworks = $demo_extra_frameworks[demo] || []
```
pepone: Don't need this frameworks for OSX demos

swift/Rakefile
```diff
@@ -140,6 +141,8 @@ def create_uikit_targets(project, demo, aggregate_target)
 
     target_add_carthage_framework(target, :ios, "Ice.framework", true)
     target_add_carthage_framework(target, :ios, "PromiseKit.framework", true)
+    target_add_carthage_framework(target, :ios, "MessageKit.framework", true)
+    target_add_carthage_framework(target, :ios, "InputBarAccessoryView.framework", true)
```
pepone: You should move this to `$demo_extra_frameworks` and only add them to `iOS/chat`

```
"iOS/chat" => ["Glacier2.framework", 
                "MessageKit.framework", 
                "InputBarAccessoryView.framework"],
```

swift/iOS/Chat/ChatLayoutController.swift
ReeceHumphreys: Sorry about that, forgot to remove the default copyright header

externl:
When I type in text into the password box then click back into the view to dismiss the keyboard, the text in the password box disappears.

externl:
My messages don't appear on the right side as expected.

IMG_0623

PR #50, opened 2019-07-04
Add C# Ice/partial demo that shows how to use partial classes with Ice

pepone:


csharp/Ice/partial/TodoItem.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        //
+        // All generated constructors call the ice_initialize partial method after
+        // initializing the data members, providing an implementation for ice_initialize
+        // partial method allow us to customize the TodoItem initialization.
+        //
+        partial void ice_initialize()
+        {
+            mId = nextId++;
+            mStatus = Status.Pending;
```
bernardnormier: I would move this to a data member with a default value in the Slice definition.

csharp/Ice/partial/config.client
```diff
@@ -0,0 +1,10 @@
+#
+# The client reads this property to create the reference to the
+# "ContactDB" object in the server.
```
bernardnormier: Should be "todolist" object?

csharp/Ice/partial/config.server
```diff
@@ -0,0 +1,11 @@
+#
+# The server creates one single object adapter with the name
+# "ContactDB". The following line sets the endpoints for this
+# adapter.
```
bernardnormier: todolist object again?

pepone:
Updated the demo as suggested

PR #55, opened 2019-07-24
Added (dispatch) Interceptor demo for Java, C#, and C++11

InsertCreativityHere:


bernardnormier:
This demo should be merged into 3.7.3-rc, since it depends on the fix to https://github.com/zeroc-ice/ice/issues/435.

InsertCreativityHere:
I ended up making the demos fully interactive after all. Everything else should be fixed as stated, except for synchronizing the operations in ThermostatI. Getting and setting a 32bit float _should_ be atomic. At least it is in Java and C#.

pepone:
Why are we using two adapters for this demo?

csharp/Ice/interceptor/AuthenticatorI.cs
```diff
@@ -18,31 +18,25 @@ internal AuthenticatorI()
...
-        //
-        long expireTime = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() + TOKEN_LIFETIME;
-
         //
         // Create a token object, store a copy, and return it.
+        // By default tokens are valid for 1 hour after being issued.
         //
-        Token token = new Token(tokenBuilder.ToString(), username, expireTime);
+        Token token = new Token(new string(tokenValue), username,
+                                DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() + TOKEN_LIFETIME);
```
pepone: We don't use all caps even for constants `TOKEN_LIFETIME` should be `tokenLifetime` or just add a comment and the expression

cpp11/Ice/interceptor/AuthenticatorI.h
```diff
@@ -12,13 +12,14 @@
...
+    void validateToken(const std::string&);
+
 public:
 
-    AuthenticatorI();
     virtual Demo::Token getToken(std::string, std::string, const Ice::Current&) override;
-    void validateToken(const std::string&);
 
-    static const long long int TOKEN_LIFETIME = 1000 * 60 * 60;
+    static const long long TOKEN_LIFETIME = 1000 * 60 * 60;
```
pepone: can we just use a duration `std::chrono::hours(1)` in the implementation and get rid of `TOKEN_LIFETIME`

cpp11/Ice/interceptor/AuthenticatorI.cpp
```diff
@@ -0,0 +1,83 @@
...
+using namespace std;
+
+long getCurrentTimeMillis()
+{
+    auto duration = chrono::system_clock::now().time_since_epoch();
+    return chrono::duration_cast<chrono::milliseconds>(duration).count();
+}
+
+AuthenticatorI::AuthenticatorI() :
+    _rand(getCurrentTimeMillis())
```
pepone: I think it is preferable to use a random_device to seed the random number generator

``` 
 std::random_device rd;
  _rand(rd()); 
```

cpp11/Ice/interceptor/AuthenticatorI.cpp
```diff
@@ -0,0 +1,83 @@
...
+        auto token = _tokenStore[i];
+        //
+        // Check if the authenticator has issued any tokens with a matching value.
+        //
+        if(token.value == tokenValue)
+        {
+            //
+            // Delete the token if it has expired.
+            //
+            if(token.expireTime <= getCurrentTimeMillis())
```
pepone: can we instead do

```
chrono::milliseconds(token.expireTime) <= chrono::system_clock::now().time_since_epoch()
```

and remove `getCurrentTimeMillis`

pepone:
We should add `using namespace std;` in C++ demos and get rid of `std::` to be consistent what other demos.

pepone:
There is some references to "3.7.2" in C++ and C# MSBuild projects version that need to be updated to 3.7.3, you can use fixIceVersion.py from dist-utils repository to fix those.

cpp11/Ice/interceptor/Interceptor.ice
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+#pragma once
+
+["java:package:com.zeroc.demos.Ice.interceptor"]
```
pepone: You can remove the Java metadata from the C++ version

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -57,93 +57,126 @@ main(int argc, char* argv[])
...
+    string line;
+    do
     {
-        //
-        // No longer encountered since the client has authorization to use 'setTemp' now.
-        //
-        cout << "Failed to set temperature. Access denied!" << endl;
+        try
+        {
+            cout << "\n==> ";
```
pepone: It is better to add `<< flush` after output to`cout` if we are not using `<< endl`, we have previously problems with platforms that do not auto flush the output in non interactive terminals (running the demos in CI from a python script)

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -57,93 +57,126 @@ main(int argc, char* argv[])
...
+                {
+                    cout << "Enter desired temperature:" << endl;
+                    string temp;
+                    getline(cin, temp);
+                    thermostat->setTemp(stof(temp));
+                    cout << "New temperature is " << thermostat->getTemp() << endl;
+                }
+                catch(const invalid_argument&)
+                {
+                    cout << "Provided temperature is not a parsable float.";
```
pepone: missing `<< endl` here

cpp11/Ice/interceptor/AuthenticatorI.cpp
```diff
@@ -34,21 +34,16 @@ AuthenticatorI::getToken(string username, string password, const Ice::Current&)
         tokenBuilder[i] = chars.at(dist(_rand));
     }
 
-    //
-    // Create a token object, store a copy, and return it.
-    //
-    Demo::Token token;
-    token.value = string(tokenBuilder);
-    token.username = username;
+    string token = string(tokenBuilder);
```
pepone: This should use the string constructor that takes a length as the second argument, because `tokenBuilder` is not null terminated.

```
    string token(tokenBuilder, 32);

File: cpp11/Ice/interceptor/Server.cpp [hunk 9]
```diff
@@ -56,13 +56,12 @@ int main(int argc, char* argv[])
...
             //
-            unordered_set<string> securedOperations({ "setTemp" });
+            unordered_set<string> securedOperations({ "setTemp", "shutdown" });
             //
             // Create an object adapter for the thermostat.
             //
             auto thermostatAdapter = communicator->createObjectAdapter("Thermostat");
-            auto thermostat = make_shared<ThermostatI>();
-            thermostatAdapter->add(make_shared<InterceptorI>(thermostat, authenticator, securedOperations), Ice::stringToIdentity("thermostat"));
+            thermostatAdapter->add(make_shared<InterceptorI>(make_shared<ThermostatI>(), authenticator, securedOperations), Ice::stringToIdentity("thermostat"));
```
pepone: You should split this line is too long, keep it to 120

cpp11/Ice/interceptor/AuthenticatorI.cpp
```diff
@@ -57,26 +57,23 @@ AuthenticatorI::validateToken(const string& tokenValue)
 {
     lock_guard<mutex> lock(_tokenLock);
```
pepone: I think is simpler if we avoid the nested if, like:

```
    lock_guard<mutex> lock(_tokenLock);

    //
    // Check if the authenticator has issued any tokens with a matching value.
    //
    auto token = _tokenStore.find(tokenValue);
    if(token == _tokenStore.end())
    {
        throw Demo::AuthorizationException();
    }
    //
    // Delete the token if it has expired.
    //
    if(token->second <= getCurrentTimeMillis())
    {
        _tokenStore.erase(token);
        throw Demo::TokenExpiredException();
    }
```

pepone:
If the clients doesn't handle `TokenExpiredException` why bother adding one, I think We either fix the clients to handle it in some meaningful way or remove it.

cpp11/Ice/interceptor/InterceptorI.cpp
```diff
@@ -0,0 +1,43 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+#include <algorithm>
```
pepone: Seems this is no longer required

cpp11/Ice/interceptor/AuthenticatorI.cpp
```diff
@@ -0,0 +1,78 @@
...
+{
+}
+
+string
+AuthenticatorI::getToken(const Ice::Current&)
+{
+    //
+    // Generate a random 32 character long token.
+    //
+    const int tokenLength = 32;
```
bentoi: I would use a static for chars to avoid allocating the string each time. You could also get rid of the C char array by filling the token string directly:
```
    const int tokenLength = 32;
    static string chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
    uniform_int_distribution<int> dist(0, static_cast<int>(chars.length() - 1));
    string token(tokenLength, '\0');
    for(auto& c : token)
    {
        c = chars.at(dist(_rand));
    }
```

Another possibility would be to have a wider character range: `0-z`:
```
    uniform_int_distribution<> dist('0', 'z');
    string token(tokenLength, '\0');
    for(auto& c : token)
    {
        c = chars.at(dist(_rand));
    }
```

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+    int status = 0;
+
+    try
+    {
+        //
+        // CommunicatorHolder's ctor initializes an Ice communicator,
+        // and its dtor destroys this communicator.
+        //
+        Ice::CommunicatorHolder ich(argc, argv, "config.client");
+        auto communicator = ich.communicator();
```
bentoi: There's no need for a temp variable if it's only used once 😄

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+    return status;
+}
+
+void menu();
+
+int
+run(const shared_ptr<Ice::Communicator>& communicator)
+{
+    auto context = communicator->getImplicitContext();
+    auto thermostat =
```
bentoi: This fits on one line of 120 columns

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+        Ice::checkedCast<Demo::ThermostatPrx>(communicator->propertyToProxy("Thermostat.Proxy"));
+    if(!thermostat)
+    {
+        cerr << "invalid thermostat proxy" << endl;
+        return 1;
+    }
+
+    auto authenticator =
```
bentoi: This fits on one line of 120 columns

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    return status;
+}
+
+void menu();
+
+int
+run(const shared_ptr<Ice::Communicator>& communicator)
+{
+    auto context = communicator->getImplicitContext();
```
bentoi: It's better to not define variables too early as it's not clear what this variable is about and when it will be used... best if to define it below when it's needed in each of `if` block... it's actually only needed for `release-token`.

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+                    thermostat->setTemp(stof(temp));
+                    cout << "New temperature is " << thermostat->getTemp() << endl;
+                }
+                catch(const invalid_argument&)
+                {
+                    cout << "Provided temperature is not a parsable float." << endl;
+                }
+                catch(const Demo::TokenExpiredException&)
+                {
+                    cout << "Failed to set temperature. Access denied!" << endl;
```
bentoi: `Token expired!` instead of `Access denied!`?

cpp11/Ice/interceptor/Client.cpp
```diff
@@ -0,0 +1,181 @@
...
+            else if(line == "?")
+            {
+                menu();
+            }
+            else
+            {
+                cout << "Unknown command `" << line << "'" << endl;
+                menu();
+            }
+        } while(cin.good() && (line != "x"));
```
bentoi: This should go on two lines:
```
}
while(...)
```

cpp11/Ice/interceptor/Interceptor.ice
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+#pragma once
+
+module Demo
+{
+    exception AuthorizationException {}
```
bentoi: We don't use `{}` but instead:
```
exception AuthorizationException 
{
}
```

cpp11/Ice/interceptor/InterceptorI.cpp
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+#include <InterceptorI.h>
+
+using namespace std;
+
+InterceptorI::InterceptorI(shared_ptr<Ice::Object> servant, shared_ptr<AuthenticatorI> authenticator,
+                           unordered_set<string> securedOperations) :
```
bentoi: I would use a more efficient ordered set... while there isn't many operations for this demo, a customer that take example on it might have a lot more operations.

cpp11/Ice/interceptor/InterceptorI.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+bool
+InterceptorI::dispatch(Ice::Request& request)
+{
+    auto current = request.getCurrent();
+    //
+    // Check if the operation requires authorization to invoke.
+    //
+    if(_securedOperations.count(current.operation) > 0)
```
bentoi: Better use find which is more efficient with an ordered set.

cpp11/Ice/interceptor/Server.cpp
```diff
@@ -0,0 +1,80 @@
...
+        // The communicator initialization removes all Ice-related arguments from argc/argv
+        //
+        if(argc > 1)
+        {
+            cerr << argv[0] << ": too many arguments" << endl;
+            status = 1;
+        }
+        else
+        {
+            //
```
bentoi: It might not be immediately clear to the user why we create two object adapters here. I assume it's to allow the configuration of restricted endpoint to access the authenticator object? A comment to explain this would be good, here and also in the server configuration file.

csharp/Ice/interceptor/AuthenticatorI.cs
```diff
@@ -0,0 +1,75 @@
...
+        for(var i = 0; i < bytes.Length; i++)
+        {
+            tokenValue[i] = chars[bytes[i] % chars.Length];
+        }
+
+        string token = new string(tokenValue);
+        //
+        // By default tokens are valid for 1 hour after being issued.
+        //
+        long expireTime = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() + (1000 * 60 * 60);
```
bentoi: Like for C++, it would be good to check if there's a better way to compare times and add offsets in C#.

cpp11/Ice/interceptor/InterceptorI.cpp
bentoi: Right, find is what we want here, we don't need the count and I imagine count will always be a little more expensive since it requires iterating over the elements to count them.

cpp11/Ice/interceptor/Client.cpp
bentoi: `get-token` doesn't really need a convenience context variable, it can just use `communicator->getImplicitContext()->put("accessToken", token);`

`getImplicitContext` is cheap and IMO it's best to privilege clarity demos. Defining `context` where it's actually used is better than defining it somewhere where it's no longer visible at the point of use.

PR #118, opened 2020-09-16
Update C# demos to use the 4.0 C# mapping

pepone:
This PR update C# demos to use the new 4.0 mapping

- The demo projects use separate source directories for client and server
- The Client/Server entry are defined as top-level applications no more Program class with static main (new in .NET 5)
- The configuration files are mostly replace by App.config settings
- Code style updates

There is more cleanup to be done, this is just a first pass to get things working with 4.0


bernardnormier (approved):
Looks good!

PR #119, opened 2020-10-20
Simplifying hello in csharp

RedTaylor:


csharp/Ice/hello/Hello.ice
```diff
@@ -8,7 +8,7 @@ module Demo
 {
     interface Hello
     {
-        idempotent void sayHello(int delay);
-        void shutdown();
+        void sayHello();
+        string sayHi();
```
externl, line 12: I think these method names are confusing. Maybe `sayHello` and `receiveHello`? Maybe we could have a single operation that takes a `struct Person` and returns a hello message `"Hello, FirstName LastName"`?

csharp/Ice/hello/server/Hello.cs
```diff
@@ -7,19 +7,13 @@ namespace Demo
 {
     public class Hello : IHello
     {
-        public void SayHello(int delay, Current current)
+        public string SayHi(Current current, global::System.Threading.CancellationToken cancel)
```
pepone, line 10: We should add a using `using System.Threading` and simplify the names in the operation signatures

```suggestion
        public string SayHi(Current current, CancellationToken cancel)
```

csharp/Ice/hello/client/Client.cs
```diff
@@ -9,106 +9,8 @@
...
-    T: set a timeout
-    P: set a server delay
-    S: switch secure mode on/off
-    s: shutdown server
-    x: exit
-    ?: help
-");
-}
+twoway.SayHello();
+Console.Out.WriteLine(twoway.SayHi());
```
externl, line 16: Newline missing at the end of the file

csharp/Ice/hello/client/Client.cs
```diff
@@ -9,106 +9,8 @@
 // The new communicator is automatically destroyed (disposed) at the end of the using statement
 using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
 
-// The communicator initialization removes all Ice-related arguments from args
-if (args.Length > 0)
-{
-    throw new ArgumentException("too many arguments");
-}
-
 IHelloPrx twoway = communicator.GetPropertyAsProxy("Hello.Proxy", IHelloPrx.Factory) ??
```
pepone, line 12: I would rename this proxy as `hello`

csharp/Ice/hello/Hello.ice
pepone: we use `idempotent string getGreeting();` in `IceDiscovery\replication\Hello.ice`

csharp/Ice/hello/server/Server.cs
```diff
@@ -11,15 +11,6 @@
 
     using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
```
pepone, line 12: we can remove the try/catch and return status if we want this to be more simple

csharp/Ice/hello/Hello.ice
InsertCreativityHere: I think the method names could use improvement, but I think it's best to leave off `idempotent` here. Technically it's correct to mark it idempotent, but this is supposed to be a bare-bones demo for people who are completely new to ice.

Adding idempotent (which is not a normal keyword in any other language I know of) adds complexity here and isn't a useful concept to the user when they're just starting out with ice I think.

csharp/Ice/hello/Hello.ice
```diff
@@ -9,6 +9,6 @@ module Demo
     interface Hello
     {
         void sayHello();
-        string sayHi();
+        string RecieveResponse();
```
externl, line 12: This should be: 
```suggestion
        string recieveResponse();
```

csharp/Ice/hello/client/Client.cs
```diff
@@ -13,4 +13,4 @@
     throw new ArgumentException("invalid proxy");
 
 twoway.SayHello();
-Console.Out.WriteLine(twoway.RecieveResponse());
+Console.Out.WriteLine(twoway.recieveResponse());
```
externl, line 16: This part should be `PascalCase` :)

csharp/Ice/hello/server/Hello.cs
```diff
@@ -7,11 +7,11 @@ namespace Demo
 {
     public class Hello : IHello
     {
-        public string RecieveResponse(Current current, global::System.Threading.CancellationToken cancel)
+        public string recieveResponse(Current current, CancellationToken cancel)
```
externl, line 10: Here too

csharp/Ice/hello/Hello.ice
```diff
@@ -8,7 +8,7 @@ module Demo
 {
     interface Hello
     {
-        idempotent void sayHello(int delay);
-        void shutdown();
+        void sayHello();
+        string recieveResponse();
```
bernardnormier, line 12: receive is spelled `receive`. And naturally we should not have such an operation. sayHello should return a string - the response - not void. `sayHello` should also take a parameter, like:

```
string sayHello(string greeting);
```

csharp/Ice/hello/server/Server.cs
```diff
@@ -11,15 +11,6 @@
...
 
-    // Destroy the communicator on Ctrl+C or Ctrl+Break
-    Console.CancelKeyPress += (sender, eventArgs) => communicator.DisposeAsync();
```
bernardnormier, line 15: We should keep this.

csharp/Ice/hello/server/Hello.cs
```diff
@@ -1,19 +1,20 @@
...
 namespace Demo
 {
     public class Hello : IHello
     {
-        public string recieveResponse(Current current, CancellationToken cancel)
-        {
-            return "Hi there!";
-        }
-        public void SayHello(Current current, CancellationToken cancel)
+        public string SayHello(Current current, CancellationToken cancel, string[] greeting)
```
pepone, line 11: `greeting` should be the first argument here, `current` and `cancel` are the last arguments for all dispatch operations and follow the user-defined arguments.

This doesn't match the definition of `sayHello` in Slice

csharp/Ice/hello/server/Hello.cs
```diff
@@ -1,19 +1,20 @@
...
         {
             Console.Out.WriteLine("Hello World!");
+            if (greeting.Length > 0)
+            {
+            return greeting;
```
pepone, line 16: this is not correctly indented. You declare a function returning a string, but you are returning a string array this doesn't compile.

Not all code paths return a value, what happens if `greeting.Length == 0`

csharp/Ice/hello/client/Client.cs
pepone: remember to rename this proxy as `hello`

RedTaylor:
Changes on this branch have been added to "implament-cancellation-tokens" so this seems unnecessary.

PR #121, opened 2020-10-29
Master

zhangbo17964:
learning how to keep connection alive


pepone:
@zhangbo17964 what are you trying to do? we don't want to merge master into 3.7, we will eventually cut the 4.0 branch from the master branch but there is still much work to do.

PR #122, opened 2020-12-09
Update Csharp demos to use async API

RedTaylor:
Editing to use Async


csharp/Ice/latency/client/Client.cs
```diff
@@ -7,7 +7,7 @@
 using ZeroC.Ice;
 
 // using statement - communicator is automatically destroyed at the end of this statement
```
bernardnormier, line 9: The comment needs to be fixed. For example:
await using - calls automatically DisposeAsync on communicator why it goes out of scope

csharp/Ice/latency/client/Client.cs
```diff
@@ -7,7 +7,7 @@
...
-using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
+await using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
 
 // Destroy the communicator on Ctrl+C or Ctrl+Break
 Console.CancelKeyPress += (sender, eventArgs) => communicator.DisposeAsync();
```
bernardnormier, line 13: Should be:
```
   // Calls DisposeAsync on Ctrl+C or Ctrl+Break, but does not wait until DisposeAsync completes.
   Console.CancelKeyPress += async (sender, eventArgs) => await communicator.DisposeAsync();
```

csharp/Ice/latency/server/Server.cs
```diff
@@ -8,7 +8,7 @@
 using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
```
bernardnormier, line 8: Should be await using like in the client. Also please fix comment like in the client.

csharp/Ice/latency/server/Server.cs
```diff
@@ -8,7 +8,7 @@
...
 
 // Destroy the communicator on Ctrl+C or Ctrl+Break
```
bernardnormier, line 10: The comment needs to be fixed, e.g.:
```
// Initiates communicator shutdown on Ctrl+C or Ctrl+Break, but does not wait for shutdown to complete.
```

Except for dispose vs shutdown, it's just like in the client, with the same "start but don't wait for completion" behavior. We use async/await in the client to avoid a warning because DisposeAsync returns a value task. ShutdownAsync returns a Task so no warning and no need for async/await.

bernardnormier: Then further down, it should be:
```
await adapter.ActivateAsync();
await communicator.WaitForShudownAync();
```

csharp/Ice/latency/client/Client.cs
```diff
@@ -7,7 +7,7 @@
 using ZeroC.Ice;
 
 // using statement - communicator is automatically destroyed at the end of this statement
-using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
+await using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
```
bernardnormier, line 10: After the communicator creation, we need to activate this communicator (it's a recent change):

```
// Activates the communicator. In a simple demo like this one, this activation typically does nothing. It is however
// recommended to always activate a communicator.
await communicator.ActivateAsync();
```

csharp/Ice/async/client/Client.cs
```diff
@@ -8,10 +8,11 @@
 try
 {
     // using statement - communicator is automatically destroyed at the end of this statement
-    using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
+    await using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
```
bernardnormier, line 11: Please fix the comment.

csharp/Ice/async/client/Client.cs
```diff
@@ -8,10 +8,11 @@
...
+    await communicator.ActivateAsync();
 
     // Destroy the communicator on Ctrl+C or Ctrl+Break
-    Console.CancelKeyPress += (sender, eventArgs) => communicator.Dispose();
+    Console.CancelKeyPress += async (sender, eventArgs) => await communicator.DestroyAsync();
```
bernardnormier, line 15: No need to async / await. And please fix the comment to
```
// Initiate the destruction of the communicator on ...
```

csharp/Ice/async/server/Server.cs
```diff
@@ -24,7 +25,7 @@
     {
         eventArgs.Cancel = true;
         workQueue.Destroy();
-        communicator.Dispose();
+        communicator.ShutdownAsync();
```
bernardnormier, line 28: Should be:
```
_ = communicator.ShutdownAsync();
```

and TODO: what about this cancel in the comment?

csharp/Ice/asyncInvocation/client/Client.cs
```diff
@@ -9,10 +9,11 @@
 try
 {
     // using statement - communicator is automatically destroyed at the end of this statement
-    using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
+    await using var communicator = new Communicator(ref args, ConfigurationManager.AppSettings);
+    await communicator.ActivateAsync();
 
     // Destroy the communicator on Ctrl+C or Ctrl+Break
-    Console.CancelKeyPress += (sender, eventArgs) => communicator.DisposeAsync();
+    Console.CancelKeyPress += async (sender, eventArgs) => await communicator.DisposeAsync();
```
bernardnormier, line 16: No async/await.

csharp/Ice/interceptor/client/Client.cs
```diff
@@ -34,7 +35,7 @@
     {
         case "1":
         {
-            Console.Out.WriteLine($"Current temperature is {thermostat.GetTemp()}");
+            await Console.Out.WriteLineAsync($"Current temperature is {thermostat.GetTemp()}");
```
bernardnormier, line 38: We keep the plain synchronous WriteLine.

PR #126, opened 2020-12-30
Simplify async demos to use asyncio

bernardnormier:
This PR simplifies the async and asyncInvocation demos by switching to await and asyncio.

### DIFF
--- a/python/Ice/async/client.py
+++ b/python/Ice/async/client.py
@@ -5,7 +5,6 @@
 
 import asyncio
 import sys
-import traceback
 import Ice
 
 Ice.loadSlice('Hello.ice')
@@ -91,5 +90,6 @@ s: shutdown server
 x: exit
 ?: help
 """)
+
 
 sys.exit(asyncio.run(main()))
--- a/python/Ice/async/client.py
+++ b/python/Ice/async/client.py
@@ -3,6 +3,7 @@
 # Copyright (c) ZeroC, Inc. All rights reserved.
 #
 
+import asyncio
 import sys
 import traceback
 import Ice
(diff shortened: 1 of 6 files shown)


pepone (approved):
Looks good

pepone:
If you merge this into 3.7 don't forget to cherry-pick to 3.7.5-rc too, or just update the PR to merge in 3.7.5-rc

python/Ice/async/client.py
```diff
@@ -3,55 +3,82 @@
...
+
+
+async def getNextChar():
+    # run blocking IO with default executor
+    c = await asyncio.get_running_loop().run_in_executor(None, sys.stdin.readline)
+    return c.strip()
+
+
+async def slowSayHello(hello):
+    await Ice.wrap_future(hello.sayHelloAsync(5000))
```
bentoi, line 80: Why don't we catch the exception directly here instead of having a "done" set for done tasks?

python/Ice/async/client.py
```diff
@@ -3,55 +3,82 @@
...
-                menu()
-            else:
-                print("unknown command `" + c + "'")
-                menu()
-        except Ice.Exception as ex:
-            print(ex)
+            nextCharTask = asyncio.create_task(getNextChar())
+            tasks.add(nextCharTask)
+            while nextCharTask not in done:
+                done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
bentoi, line 41: It's not clear to me why we maintain a tasks set here instead of just directly awaiting on `getNextChar` (or just directly call await on `loop.run_in_executor(None, sys.stdin.readline)`) or await on the `sayHelloAsync` calls.

bentoi (approved):
Should we consider using a dispatch instead of `wrap_future`? Futures returned by Ice async calls would get completed through the dispatcher which would call `loop.call_soon_threadsafe` to ensure that the continuation of the future is ran on event loop. 

We could also consider adding a loop parameter to the communicator initialization (if set this would setup a dispatcher to run calls on the loop).

python/Ice/async/client.py
bernardnormier: Because this done processing works for all tasks. For example, without it, if you try to shutdown a server that is already shutdown, you'd get:
```
==> s
Future exception was never retrieved
future: <Future finished exception=exception ::Ice::ConnectionRefusedException
{
    error = 61
}>
Ice.ConnectionRefusedException: Ice.ConnectionRefusedException:
```

bernardnormier: However, the `else:` was not correct: we should do this processing all the time.

python/Ice/async/client.py
bernardnormier: I don't understand your comment: we simply launch a number of tasks and await them, i.e. let them run "in parallel".

bernardnormier:
> Should we consider using a dispatch instead of `wrap_future`? Futures returned by Ice async calls would get completed through the dispatcher which would call `loop.call_soon_threadsafe` to ensure that the continuation of the future is ran on event loop.
> 
> We could also consider adding a loop parameter to the communicator initialization (if set this would setup a dispatcher to run calls on the loop).

Your comment is not clear to me.

bernardnormier:
> If you merge this into 3.7 don't forget to cherry-pick to 3.7.5-rc too, or just update the PR to merge in 3.7.5-rc

Merged to 3.7.5-rc.

bentoi:
My idea was to do the same as the C# async demo... something along these lines:
```
async def main():

    # Ice.initialize returns an initialized Ice communicator; the communicator is destroyed once it goes out of scope.
    with Ice.initialize(sys.argv, "config.client") as communicator:

        # The communicator initialization removes all Ice-related arguments from sys.argv
        if len(sys.argv) > 1:
            print(sys.argv[0] + ": too many arguments")
            return 1

        try:
            hello = Demo.HelloPrx.uncheckedCast(communicator.propertyToProxy('Hello.Proxy'))
        except Ice.Exception as ex:
            print("invalid proxy:", ex)
            return 1

        menu()

        while True:
            sys.stdout.write("==> ")
            sys.stdout.flush()
            c = await asyncio.get_event_loop().run_in_executor(None, sys.stdin.readline)
            c = c.strip()

            if c == 'i':
                await Ice.wrap_future(hello.sayHelloAsync(0))
            elif c == 'd':
                asyncio.create_task(slowSayHello(hello))
            elif c == 's':
                await Ice.wrap_future(hello.shutdownAsync())
            elif c == 'x':
                return 0
            elif c == '?':
                menu()
            else:
                print(f"unknown command {c!r}")
                menu()

async def slowSayHello(hello):
    try:
        await Ice.wrap_future(hello.sayHelloAsync(5000))
        print("slow hello completed")
    except Ice.Exception as ex:
        print("task exception:", ex)
    except RuntimeError as ex:
        print("task runtime error:", ex)
```

While `sayHelloAsync(0)` and `shutdown()` won't run in parallel, I don't think that's a big deal, the demo is a lot simpler like this.

Another step could be to get rid of the `Ice.wrap_future` calls by using a dispatcher that ensures that AMI continuations are ran from the event loop. I tried this but for some reasons it doesn't work, I'm getting a `RuntimeError: Task got bad yield: <Ice.InvocationFuture object at 0x10a374fd0>` from the yield in `IceFuture.__await__` (defined in `IceFuture.py`).

PR #127, opened 2021-01-06
Python Ice/async demo improvements

bentoi:


python/Ice/async/client.py
```diff
@@ -22,62 +22,51 @@ async def main():
...
+                elif c == 'd':
+                    # create asyncio task from coroutine returned by slowSayHello
+                    asyncio.create_task(slowSayHello(hello))
+                elif c == 's':
+                    # create asyncio future from Ice future
+                    await Ice.wrap_future(hello.shutdownAsync())
+                elif c == '?':
+                    menu()
+                else:
+                    print(f"unknown command {c!r}")
```
bernardnormier, line 53: The 'x' handling was lost.

Then, when we press 'x', should we wait for all outstanding 'd' tasks to complete? Or cancel them? Currently it just exits immediately.

python/Ice/async/client.py
```diff
@@ -22,62 +22,51 @@ async def main():
...
-
-
-async def getNextChar():
-    # run blocking IO with default executor
-    c = await asyncio.get_running_loop().run_in_executor(None, sys.stdin.readline)
-    return c.strip()
+        c = None
+        while c != 'x':
+            try:
+                sys.stdout.write("==> ")
```
bernardnormier, line 35: Should we instead
```
print("==> ", end='', flush=True)
```
?

bernardnormier (approved):

python/Ice/async/client.py
bentoi: I'm not sure it makes a big difference given that we're exiting but I added back the cancellation of pending tasks. I also added code to ensure we wait for these tasks to complete and handle the CancelledError from the coroutine.

python/Ice/async/client.py
bentoi: Either if fine with me as long as it still works with our expect demo scripts. If we decide to change this, it's best to do this for all demos separately from this PR.

python/Ice/async/client.py
```diff
@@ -15,69 +15,71 @@ async def main():
...
-                    except RuntimeError as ex:
-                        print("task runtime error:", ex)
-
-
-async def getNextChar():
-    # run blocking IO with default executor
-    c = await asyncio.get_running_loop().run_in_executor(None, sys.stdin.readline)
-    return c.strip()
+                        await asyncio.wait(tasks)
+                    pass
```
bernardnormier, line 59: Do you need this pass?

bentoi: Nope, it's a left over.

PR #128, opened 2021-01-06
Rework Python coroutine demo

bernardnormier:
Also added tox.ini file to ice-demos.


python/Ice/coroutine/server.py
```diff
@@ -3,93 +3,75 @@
...
+    elif n == 1:
+        return 0
+    elif n == 2:
+        return 1
+    else:
+        return fibonacci(n - 2) + fibonacci(n - 1)
+
+
+# The servant for the back-end. It would typically live in a separate back-end server.
+class FibonacciBackEndI(Demo.Fibonacci):
```
externl, line 28: Should this be `FibonacciBackEndI(Demo.FibonacciBackEnd):`?

externl (commented):
Could we just use the executor directly in `FibonacciI` instead of having to call another proxy?

python/Ice/coroutine/server.py
bernardnormier: Yes, it should be!

bernardnormier:
> Could we just use the executor directly in `FibonacciI` instead of having to call another proxy?

Of course, but the whole point of the demo is to show a dispatch calling a proxy.

pepone (approved):
Looks good

python/Ice/coroutine/client.py
```diff
@@ -5,40 +5,70 @@
...
 def run(communicator):
+    fibonacci = Demo.FibonacciPrx.uncheckedCast(communicator.propertyToProxy('Fibonacci.Proxy'))
 
-    fetcher = Demo.FetcherPrx.checkedCast(communicator.propertyToProxy('Fetcher.Proxy'))
-    if not fetcher:
-        print("invalid proxy")
-        sys.exit(1)
+    try:
+        v = fibonacci.compute(-1)
+        print(f"fibonacci(-1) = {v:!r}")  # unexpected
```
bentoi, line 20: I'd add or replace the print with `assert False`

python/Ice/coroutine/server.py
```diff
@@ -3,93 +3,75 @@
...
-    adapter = communicator.createObjectAdapter("Fetcher")
-    adapter.add(FetcherI(loop), Ice.stringToIdentity("fetcher"))
-    adapter.activate()
-
-    #
-    # Run the asyncio event loop until stop() is called.
-    #
-    loop.run_forever()
-    loop.close()
+# The actual recursive fibonacci implementation. It's written in Python so it executes with the GIL locked.
```
bentoi, line 15: I find the comment about the GIL here a bit surprising. I would move it to the executor submit call and explain CPython limitation caused by the GIL which requires to use separate processes to perform computations on several cores.

bentoi (approved):
Looks good but I wonder if it wouldn't be clearer to have 3 processes rather than 2, the client, the frontend and the backend processes. It would be a bit more complex however.

externl (approved):

python/Ice/coroutine/server.py
bernardnormier: Fixed.

python/Ice/coroutine/client.py
bernardnormier: Fixed.

python/Ice/coroutine/server.py
bernardnormier: Fixed.

PR #142, opened 2021-06-03
Update Swift demos to use XCFrameworks

pepone:
This PR updates Swift demos to use XCFramework


swift/Cartfile
```diff
@@ -1,2 +1,2 @@
 github "MessageKit/MessageKit" ~> 3.0
-github "zeroc-ice/ice" ~> 3.7
+github "zeroc-ice/ice" "3.7"
```
externl, line 2: what's the difference between `~> 3.7` and `"3.7"`

pepone: yes see https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#version-requirement

swift/Rakefile
```diff
@@ -212,8 +212,8 @@ def target_add_slice2swift_build_rule(project, target, prefix = nil)
     rule.file_type = "pattern.proxy"
 
     script = <<~SCRIPT
-        if [ -f "${ICE_HOME-unset}/cpp/bin/slice2swift" ]; then
-            SLICE2SWIFT="$ICE_HOME/cpp/bin/slice2swift"
+        if [ -f "${ICE_HOME}/bin/slice2swift" ]; then
```
externl, line 215: I think we should keep it as `${ICE_HOME-unset}` in case ICE_HOME is not set. We don't want to accidentally hit `/bin/slice2swift`

externl (approved):
Looks good

swift/Rakefile
pepone: what is `-unset` for, is this a special keyword?

externl: It can be anything, it's just to provide a default in case the variable is not set not yet when you expect it to be. For example, it's often used to prevent against something like `rm -rf ${TYPOO}/`.

PR #147, opened 2022-02-11
Converted iOS demos to SwiftUI and updated folder structure

### LINKED ISSUE #95: Add Swift UI demos
We should update iOS graphical demos to use Swift UI, maybe we can also support macOS with those

ReeceHumphreys:
The goal of this PR is to update the iOS demos to SwiftUI. Not only does this allow for the demos to be more modern, but it also comes with the benefits of automatic dark mode, support for macOS catalyst, and easier to manage views.

Additionally, the following restructuring was performed:
- The hello demo was split into two versions, one that uses Ice discovery and a version that does not.
- The library demo was not illustrating anything that the chat or hello did not already accomplish, so it was removed
- The chat demo was moved to the root of the swift directory to match the other languages.

Closes #95


ReeceHumphreys:
@pepone Would you be able to help with the Rakefile for this PR?

ReeceHumphreys:
There is a missing commit for the hellos from when I split them into two. Will be adding that commit. Currently it is just two copies of the Unified hello demo before the split.

swift/Ice/helloUI/README.md
```diff
@@ -0,0 +1,43 @@
+This demo illustrates how to integrate Ice with a simple graphical
+application.
```
pepone, line 2: Maybe mention SwiftUI directly.

```
This demo illustrates how to write a SwiftUI application that invokes
ordinary (twoway) operations, as well as how to make [oneway][1],
[datagram][2], [secure][3], and [batched][4] invocations.
```

swift/Chat/views/MessageView.swift
```diff
@@ -0,0 +1,275 @@
...
+    func inputBar(_ inputBar: InputBarAccessoryView, didPressSendButtonWith _: String) {
+        let text = inputBar.inputTextView.text ?? ""
+        inputBar.inputTextView.text = String()
+        inputBar.invalidatePlugins()
+        inputBar.sendButton.startAnimating()
+        inputBar.inputTextView.placeholder = "Sending"
+        DispatchQueue.global(qos: .default).async {
+            self.client.session?.sendAsync(text).catch {
+                print($0)
+                // self.exception(err: $0)
```
pepone, line 217: Why is this commented out, how do we handle the send error? What about the print statement is it temporary?

pepone (commented):
Just a few comments, still need to build and run the demos...

swift/LoginView.swift
```diff
@@ -0,0 +1,64 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+import SwiftUI
```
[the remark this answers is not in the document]
ReeceHumphreys: Yes seems that when I was moving some files around duplicates got created. Will remove them.

swift/Chat/Client.swift
```diff
@@ -0,0 +1,181 @@
...
+                              category: category)
+                    loginViewModel.connecting = false
+
+                    //
+                    // The communicator is now owned by the ChatController.
+                    //
+                    if loginViewModel.rememberMe == true {
+                        UserDefaults.standard.set(loginViewModel.username, forKey: Configuration.Keys.user.rawValue)
+                        UserDefaults.standard.set(loginViewModel.password, forKey: Configuration.Keys.password.rawValue)
+                    }
```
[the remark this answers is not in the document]
ReeceHumphreys: Just so I understand, you are suggesting we remove the remember me features as a whole since user defaults is insecure?

swift/Chat/views/MessageView.swift
ReeceHumphreys: Yes the print statement was temporary. Just added proper displaying of error. Will commit changes that address other comments shortly too!

InsertCreativityHere:
It looks like the swift demo's README (`swift/README.md`) needs to be updated.

On line 16, it mentions the `iOS` directory, which this PR deletes.
And it doesn't mention the new `Chat` directory which this PR adds.

You should remove the message for `iOS` and add a message for `Chat`.

swift/Chat/README.md
```diff
@@ -1,4 +1,4 @@
-This demo is an iOS client for the [ZeroC Chat Demo][1]. It connects to a
+This demo is an SwiftUI client for the [ZeroC Chat Demo][1]. It connects to a
```
InsertCreativityHere, line 1:
```suggestion
This demo is a SwiftUI client for the [ZeroC Chat Demo][1]. It connects to a
```

swift/Ice/helloUI/Client.swift
```diff
@@ -0,0 +1,212 @@
...
+        } catch {
+            exception(error)
+        }
+    }
+
+    func flushBatch() {
+        if helloPrx == nil {
+            return
+        }
+        flushEnabled = true
```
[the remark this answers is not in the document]
ReeceHumphreys: You are right, flush enabled should be false. It should be enabled once users click say hello once while on a batch method.

swift/Ice/helloUI/Client.swift
```diff
@@ -0,0 +1,212 @@
...
+    func sendingRequest() {
+        helloEnabled = false
+        shutdownEnabled = false
+        statusMessage = "Sending request"
+        isSpinning = true
+    }
+
+    func requestSent() {
+        if let connection = helloPrx.ice_getCachedConnection() {
+            do {
```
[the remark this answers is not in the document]
ReeceHumphreys: Idea was to mimic business logic of prior UIKit version of these demos just with SwiftUI fixes.

swift/IceDiscovery/helloUI/views/ContentView.swift
```diff
@@ -0,0 +1,64 @@
...
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+import Darwin
+import Ice
+import PromiseKit
+import SwiftUI
+
+struct ContentView: View {
+    func loadCommunicator() -> Ice.Communicator {
```
InsertCreativityHere, line 11: I like that this logic was moved into it's own function, but I think it belongs in Client, instead of ContentView.
I'm guessing ContentView is like the 'main' view?

swift/IceDiscovery/helloUI/Client.swift
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    func exception(_ err: Error) {
+        error = err
+        showingError = true
+    }
+
+    func sayHello() {
+        do {
+            if helloPrx == nil {
```
bernardnormier, line 45: helloPrx (as declared) does not allow nil values.

swift/Ice/helloUI/Client.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    func exception(_ err: Error) {
+        error = err
+        showingError = true
+    }
+
+    func sayHello() {
+        do {
+            if helloPrx == nil {
```
bernardnormier, line 45: helloPrx does not allow nil values.

swift/Ice/helloUI/Client.swift
```diff
@@ -0,0 +1,208 @@
...
+            exception(error)
+            ready()
+        }
+    }
+
+    func updateProxy() throws {
+        let mode = proxySettings.deliveryMode
+
+        if proxySettings.hostname == "" {
+            helloPrx = nil
```
bernardnormier, line 129: You should not assign nil into a non-optional variable.

swift/IceDiscovery/helloUI/Client.swift
```diff
@@ -0,0 +1,199 @@
...
+                    }
+                }.done {
+                    response = true
+                    self.ready()
+                }.catch { error in
+                    response = true
+                    self.exception(error)
+                }
+            } else {
+                try helloPrx.sayHello(delay)
```
bernardnormier, line 68: I think the logic here is it's a batch request, so we "send" it synchronously because we actually only enqueue it locally. Would be nice to add a comment.

swift/Ice/helloUI/Client.swift
```diff
@@ -0,0 +1,208 @@
...
+
+    func sendingRequest() {
+        helloEnabled = false
+        shutdownEnabled = false
+        statusMessage = "Sending request"
+        isSpinning = true
+    }
+
+    func requestSent() {
+            if let connection = helloPrx.ice_getCachedConnection() {
```
bernardnormier, line 172: I think this hostname overwriting doesn't make sense for helloUI: you've entered hostname, obviously that's the hostname the Client is going to use to connect to the server. Here we're just replacing hostname with the remote IP address of the connection established with the server.

swift/IceDiscovery/helloUI/Client.swift
```diff
@@ -0,0 +1,199 @@
...
+        helloEnabled = false
+        shutdownEnabled = false
+        statusMessage = "Sending request"
+        isSpinning = true
+    }
+
+    func requestSent() {
+        if let connection = helloPrx.ice_getCachedConnection() {
+            do {
+                // Loop through the connection informations until we find an IPConnectionInfo class.
```
bernardnormier, line 164: In this discovery demo, hostname is read-only and we show the hostname (but not port number) of the connection used to send the request. 

I think we could do clearer, and just print the connection (and rename hostname to connection) to show which connection was used to send the request.

swift/Chat/Client.swift
```diff
@@ -0,0 +1,170 @@
...
+        // to modify the UI in a thread safe manner.
+        //
+        return try DispatchQueue.main.sync {
+            try self.servantDisp.dispatch(request: request, current: current)
+        }
+    }
+}
+
+class Client: ChatRoomCallback, ObservableObject {
+    var communicator: Ice.Communicator?
```
bernardnormier, line 27: Make some or all these fields private?

bernardnormier (commented):
Good start!

swift/Chat/Client.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    // MARK: - Client management functions
+
+    func setup(communicator: Ice.Communicator,
+               session: ChatSessionPrx,
+               acmTimeout: Int32,
+               router: Glacier2.RouterPrx,
+               category: String) throws
+    {
+        self.communicator = communicator
```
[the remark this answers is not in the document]
pepone: Maybe is simpler to remove this `setup` method and move the code to the caller.

swift/Chat/models/LoginViewModel.swift
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return true
+    }
+
+    // Validation methods
+    private func isUsernameValid() -> Bool {
+        return username != ""
+    }
+
+    private func isPasswordValid() -> Bool {
```
pepone, line 36: What is the point of this method, if it always return true?

swift/Chat/views/LoginView.swift
```diff
@@ -0,0 +1,85 @@
+//
+// Copyright (c) ZeroC, Inc. All rights reserved.
+//
+
+import SwiftUI
+
+struct LoginView: View {
+    @ObservedObject var client = Client()
+    @State var showingUsers = false
```
pepone, line 9: Seems we don't read this variable

pepone (commented):
A few comments for the chat demo, will review the rest later.

bernardnormier (approved):
Looks good.

swift/Ice/helloUI/ProxySettings.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    func updateDeliveryMode(index: Int) -> DeliveryMode {
+        return DeliveryMode(rawValue: index) ?? .twoway
+    }
+
+    var isBatched: Bool {
+        return DeliveryMode.batchedCases.contains(deliveryMode)
+    }
+
+    enum DeliveryMode: Int, CaseIterable {
```
InsertCreativityHere, line 26: There's no point for this to have a raw type of `Int` anymore. You can _probably_ remove this.

InsertCreativityHere (approved):
I think it'd be worthwhile to clean up this demo in the (possibly immediate) future.
But as far as the scope of this PR goes: converting the demos to SwiftUI...
Looks good to me!

PR #156, opened 2022-03-04
Removed net standard 2.0

ReeceHumphreys:
Removed the build folders related to net standard 2.0
- Followed the instructions from the csharp readme for macOS and everything else still built fine.
- Not sure if I did indeed remove everything properly as the csharp build system is confusing, so let me know if something doesn't make sense.


pepone (approved):
Looks good!

pepone:
The main README need updates remove the corresponding section https://github.com/ReeceHumphreys/ice-demos/tree/removing-netstandard/csharp#building-and-running-the-demos-with-net-core-on-windows-linux-and-macos

csharp/C# NET Core demos.sln
```diff
@@ -1,1253 +0,0 @@
-Microsoft Visual Studio Solution File, Format Version 12.00
```
externl, line 1: Is this the right file to delete?

pepone: Yes that is fine

externl (approved):
Looks good!

PR #161, opened 2022-04-11
Removed Objective-C UI demos

ReeceHumphreys:


InsertCreativityHere:
You should check through the README's to remove any references to removed demos, like:
https://github.com/ReeceHumphreys/ice-demos/blob/removing-objc-ui-demos/objective-c/Ice/README.md
Still mentions the iOS demos. Probably others too.

InsertCreativityHere (approved):
Looks good to me!

PR #172, opened 2023-02-22
Rework Android Demos

pepone:
- Update Android Hello demo for Java and Java Compat mappings
- Remove more complex demos that were outdated


java/android/hello/src/main/java/com/zeroc/hello/HelloApp.java
```diff
@@ -74,70 +70,55 @@ else if(m.what == MSG_EXCEPTION || m.what == MSG_RESPONSE)
...
-        //
-        _lock = _wifiManager.createMulticastLock("com.zeroc.demos.android.hello");
-        _lock.acquire();
-
         // SSL initialization can take some time. To avoid blocking the
         // calling thread, we perform the initialization in a separate thread.
-        new Thread(new Runnable()
-        {
-            public void run()
+        new Thread(() -> {
```
externl, line 75: We don't need to acquire the WifiManager lock ?

externl (approved):
Looks good!

java/android/hello/src/main/java/com/zeroc/hello/HelloApp.java
pepone: Yes we do, good catch!!

Issue #146, opened 2022-02-07
Avoid thrown exception type is not nothrow copy constructible [cert-err60-cpp]


ClausKlein:
I am evaluating the zeros-ice framework with **c++11** enabled mapping on **OSX**.
I use the current **ice version v3.7** and check with `clang-tidy` tools the generated code quality.

To do this, I use this `GNUmakefile`:
```make
check: compile_commands.json
	run-clang-tidy

compile_commands.json:
	bear -- $(MAKE) -f makefile -j V=1 Chat/client Chat/server
	perl -i.bak -p -e 's#-Wall#-I/usr/local/include#;' $@

.POHNY: all clean distclean
all:
	$(MAKE) -f makefile -j V=1 Chat/client Chat/server

clean:
	$(MAKE) -f makefile -j V=1 Chat/client_clean Chat/server_clean

distclean: clean
	$(MAKE) -f makefile -j V=1 Chat/client_distclean Chat/server_distclean
	$(RM) compile_commands.json *.bak

```

This is the result:
```bash
clang-tidy --use-color -p=/Users/clausklein/Workspace/cpp/ice-demos/cpp11 /Users/clausklein/Workspace/cpp/ice-demos/cpp11/Chat/client/ChatUtils.cpp
/Users/clausklein/Workspace/cpp/ice-demos/cpp11/Chat/client/ChatUtils.cpp:23:5: warning: use range-based for loop instead [modernize-loop-convert]
    for(unsigned int count = 0; count < sizeof(htmlEntities) / sizeof(htmlEntities[0]); ++count)
    ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       (const auto & htmlEntitie : htmlEntities)
64088 warnings generated.
Suppressed 64087 warnings (64087 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
clang-tidy --use-color -p=/Users/clausklein/Workspace/cpp/ice-demos/cpp11 /Users/clausklein/Workspace/cpp/ice-demos/cpp11/Chat/server/ChatUtils.cpp
Chat/server/ChatUtils.cpp:49:15: warning: thrown exception type is not nothrow copy constructible [cert-err60-cpp]
        throw Chat::InvalidMessageException(os.str());
              ^
[...]

PR #178, opened 2023-03-09
Fix issues reported by clang-tidy

pepone:
This PR updates C++11 demos, to fix warnings reported by clang-tidy, the majority are about const correctness, for places where we forget to declare a variable const. There are a few other things like `std::move` usage and some other minor issues.

see also #146

### DIFF
--- a/cpp98/IceGrid/customLocator/Client.cpp
+++ b/cpp98/IceGrid/customLocator/Client.cpp
@@ -109,10 +109,6 @@ run(const Ice::CommunicatorPtr& communicator, bool addContext)
             {
                 hello->sayHello();
             }
-            else if(c == 't')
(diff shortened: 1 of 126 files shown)


cpp98/IceGrid/customLocator/Client.cpp
```diff
@@ -109,10 +109,6 @@ run(const Ice::CommunicatorPtr& communicator, bool addContext)
             {
                 hello->sayHello();
             }
-            else if(c == 't')
```
externl, line 112: We should also fix this in the c++98 demo

externl (approved):
Looks good!

bernardnormier (approved):
Looks good.

PR #179, opened 2023-03-10
Handle exceptions from stoi calls

pepone:
This PR fixes `IceGrid/replication` and `IceDiscovery/replication` to correctly handle errors from `stoi`. It also updates the loop that reads the input parameters to make clear how to exit.

### DIFF
--- a/cpp11/IceDiscovery/replication/Client.cpp
+++ b/cpp11/IceDiscovery/replication/Client.cpp
@@ -74,16 +74,56 @@ run(const shared_ptr<Ice::Communicator>& communicator)
     string s;
     do
     {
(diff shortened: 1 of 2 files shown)


externl:
> It also updates the loop that reads the input parameters to make clear how to exit.

Having 3 while loops seems like a lot. Maybe we should just get rid of all these loops?

pepone:
> Having 3 while loops seems like a lot. Maybe we should just get rid of all these loops?

What about exiting and printing an error if `stoi` fails with `invalid_argument` and keeping just the main loop?

externl (approved):
Looks good!

PR #184, opened 2023-03-17
Simplify / cleanup async demo

bernardnormier:
This PR simplifies the async demo by using a CancellationTokenSource.

It also reformats the code and updates the method names to be more standard C#.

Tested with Visual Studio 2022 + .NET Framework 4.8.


csharp/Ice/async/Client.cs
```diff
@@ -15,116 +14,107 @@ public static int Main(string[] args)
...
-                    menu();
+                    case null:
+                        break;
+
+                    case "i":
+                        hello.sayHello(0);
+                        break;
+
+                    case "d":
+                        _ = HelloAsync(hello);
```
pepone, line 68: This doesn't work with Visual Studio 2015, .NET 4.5

We can do `Task.Run(() => HelloAsync(hello));`

csharp/Ice/async/WorkQueue.cs
pepone, line 1: The project still references this removed file

csharp/Ice/async/Server.cs
```diff
@@ -3,60 +3,55 @@
...
+                        };
 
-                    var adapter = communicator.createObjectAdapter("Hello");
-                    adapter.add(new HelloI(workQueue), Ice.Util.stringToIdentity("hello"));
-
-                    workQueue.Start();
-                    try
-                    {
+                        Ice.ObjectAdapter adapter = communicator.createObjectAdapter("Hello");
+                        _ = adapter.add(new HelloI(cts), Ice.Util.stringToIdentity("hello"));
```
pepone, line 35: `_` doesn't work with VS 2015

csharp/Ice/async/Server.cs
```diff
@@ -3,60 +3,55 @@
 //
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 
 public class Server
 {
-    public static int Main(string[] args)
+    public static async Task<int> Main(string[] args)
```
pepone, line 11: Visual Studio 2015 doesn't allow async main

csharp/Ice/async/Client.cs
bernardnormier: Fixed, but didn't use Task.Run.

csharp/Ice/async/WorkQueue.cs
bernardnormier: Fixed.

csharp/Ice/async/Server.cs
bernardnormier: Removed.

csharp/Ice/async/Server.cs
bernardnormier: Fixed and tested with VS 2015.

bernardnormier:
Tested with VS 2015 Update 3.

bentoi (approved):

csharp/Ice/async/Client.cs
```diff
@@ -15,121 +14,107 @@ public static int Main(string[] args)
 
         try
         {
-            //
-            // using statement - communicator is automatically destroyed
-            // at the end of this statement
-            //
-            using(var communicator = Ice.Util.initialize(ref args, "config.client"))
+            using (Ice.Communicator communicator = Ice.Util.initialize(ref args, "config.client"))
```
externl, line 17: Why get rid of the comment?

csharp/Ice/async/README.md
```diff
@@ -19,12 +19,9 @@ client
 > dotnet client.dll
 > ```
 
-The demo invocation can either have a short response time or require a
-significant amount of time to complete. For the long running request
-the client uses AMI and the server uses AMD plus a worker thread to
-process the request. While a long request is processing, short
-requests are still able to be processed and more long requests can be
-queued for processing by the worker thread.
+The shutdown of an server (and its object adapter) waits for outstanding dispatches
```
externl, line 22:
```suggestion
The shutdown of a server (and its object adapter) waits for outstanding dispatches
```

csharp/Ice/async/Server.cs
```diff
@@ -12,51 +14,38 @@ public static int Main(string[] args)
...
-                        communicator.waitForShutdown();
-                    }
-                    finally
-                    {
-                        workQueue.Join();
+
+                        // cts is canceled by Ctrl+C or a shutdown request.
+                        // In modern C#, you should call: await Task.Delay(-1, cts.Token)
+                        cts.Token.WaitHandle.WaitOne();
+                        communicator.shutdown();
```
externl, line 41: I would not say "in modern C#". Or at least say that we need to do it this way for compatibility with older versions of .NET

externl (approved):

csharp/Ice/async/Client.cs
bernardnormier: What `using` is about is basic C#. It's odd to have this comment in every single server.

csharp/Ice/async/Server.cs
bernardnormier: Fixed

PR #185, opened 2023-03-17
Update/simplify asyncInvocation demo

bernardnormier:
This PR simplifies the asyncInvocation demo: it only uses await calls (nobody should use "raw" continuations) and reworks the "shutdown requested" logic in the server.


pepone (approved):
Looks good

bentoi (approved):

PR #186, opened 2023-03-19
Rework "wait for shutdown"

bernardnormier:
This PR reworks the hello Server to "wait for shutdown" using a promise/future.

It's a draft PR because if we decide to adopt this new model for cpp11 servers, we should update all the demos, not just the hello demo.

### DIFF
--- a/cpp11/Ice/hello/Server.cpp
+++ b/cpp11/Ice/hello/Server.cpp
@@ -4,6 +4,7 @@
 
 #include <Ice/Ice.h>
 #include <HelloI.h>
+#include <future>
 
 using namespace std;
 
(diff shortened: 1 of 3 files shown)


cpp11/Ice/hello/Server.cpp
```diff
@@ -48,11 +46,13 @@ int main(int argc, char* argv[])
...
             auto adapter = communicator->createObjectAdapter("Hello");
-            adapter->add(make_shared<HelloI>(), Ice::stringToIdentity("hello"));
+            adapter->add(make_shared<HelloI>(requestShutdown), Ice::stringToIdentity("hello"));
             adapter->activate();
 
-            communicator->waitForShutdown();
+            // The shutdown can be requested by Hello::shutdown or the CtrlCHandler.
+            shutdownRequested.wait();
         }
+        communicator->shutdown();
```
bernardnormier, line 55: Keep in mind `Communicator::shutdown` means "initiate shutdown". You need to call waitForShutdown if you want to wait until the shutdown is actually complete.

pepone (commented):
I find the new setup a bit more complex than the current code, what is the motivation for this change?

bentoi:
I also find the previous setup simpler.

externl:
I agree that the current code is simpler.

bernardnormier:
In the existing server boilerplate, main blocks until the communicator shutdown is completed, and various events (Ctrl+C, the shutdown operation) initiate the communicator shutdown.

In this revised version, the Ice runtime is no longer involved in this logic: main blocks until the application calls "requestShutdown", and requestShutdown does not call the Ice runtime.

PR #192, opened 2023-03-23
Do not dispose CTS until communicator is disposed

pepone:
We should dispose the CTS only after the communicator is disposed of, otherwise request cancelation doesn't work. The CTS will no throw OCE after it is disposed.

### DIFF
--- a/csharp/Ice/async/Server.cs
+++ b/csharp/Ice/async/Server.cs
@@ -9,44 +9,40 @@ public class Server
 {
     public static int Main(string[] args)
     {
-        int status = 0;
+        if (args.Length > 0)
+        {
+            Console.WriteLine("too many arguments");
+            return 1;
+        }
 
         try
         {
-            using (Ice.Communicator communicator = Ice.Util.initialize(ref args, "config.server"))
+            using (var cts = new CancellationTokenSource())
             {
-                if (args.Length > 0)
+                using (Ice.Communicator communicator = Ice.Util.initialize(ref args, "config.server"))
                 {
-                    Console.WriteLine("too many arguments");
-                    status = 1;
-                }
-                else
-                {
-                    using (var cts = new CancellationTokenSource())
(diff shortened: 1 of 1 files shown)


csharp/Ice/async/Server.cs
```diff
@@ -9,44 +9,40 @@ public class Server
 {
     public static int Main(string[] args)
     {
-        int status = 0;
+        if (args.Length > 0)
```
bernardnormier, line 12: This is not correct. It's ok and common to start the server with for example `--Ice.Trace.Protocol=1`.

bernardnormier:
I don't understand the issue this PR is trying to fix.

We dispose the CTS before the communicator is destroyed and that's totally fine.

> otherwise request cancelation doesn't work.

I am not following.

pepone:
With the current code

```
jose@ubuntu22:~/3.7.9/ice-demos/csharp/Ice/async$ dotnet client.dll
usage:
i: send immediate greeting
d: send delayed greeting
s: shutdown server
x: exit
?: help

==> d
==> s
==> 
```

Here `s` invokes shutdown, which in turn cancels the token, and the cancelation triggers its disposal. After the token is disposed it no longer cancels the `Task.Delay` as you can see from the output the request isn't canceled.

bernardnormier:
I think the issue is after:
```csharp
cts.Token.WaitHandle.WaitOne();
```

returns, we dispose the cts then destroy the communicator (in this order), there is a small chance some dispatches are still running and didn't check `cts.Token` yet.

I think a better fix would be to wait for these dispatches to complete, with:
```csharp
// existing comment
cts.Token.WaitHandle.WaitOne();

// Wait for all dispatches to complete before disposing cts 
// since the dispatches uses cts
communicator.shutdown();
communicator.waitForShutdown();
```

pepone:
> I think the issue is after

reworked as suggested.

bernardnormier (approved):

PR #194, opened 2023-06-06
Update README.md

fokman:
the evictor has not found in Manual project


externl:
Thanks for the PR @fokman - I think we can just remove this section entirely instead of commenting it out!

externl:
There's been no activity on this so I'm going to close it and open an issue to update the readme.

Issue #181, opened 2023-03-15
Server demos should show recommended server shutdown


bernardnormier:
The recommended shutdown procedure for a server is to shut it down before destroying it. In some demos,  the server is correctly shut down, for example:

https://github.com/zeroc-ice/ice-demos/blob/7d29cb7548a7c805d2a5fc612f1e7b2417e49bff/cpp11/Ice/hello/Server.cpp#L37

In other demos, the server is just destroyed. While it doesn't matter for those demos, it's not the recommended style:

https://github.com/zeroc-ice/ice-demos/blob/7d29cb7548a7c805d2a5fc612f1e7b2417e49bff/csharp/Ice/hello/Server.cs#L24

We should also consider not using `communicator.waitForShutdown()` in demos with no shutdown operation. All we need is a "WaitForCtrlC()".

bernardnormier:
Fixed in Ice 3.8.

PR #195, opened 2023-10-05
Update minimal C# demo

bernardnormier:
A small update to the minimal demo to make it more correct / cleaner.


bernardnormier:
See also #181.

externl:
Do we care about de-sycnronizing this one demo from the rest? Should we update all of the minimal demos?

pepone (approved):
Should we update other language mappings?

externl (approved):

bernardnormier:
Yes, I am in favor of updating all the other minimal demos.

PR #200, opened 2023-12-18
Upgrade C# demos to only target .netstandard2.0

pepone:
This PR upgrades the C# demos to only target .netstandard2.0, following changes on the ice main branch https://github.com/zeroc-ice/ice/pull/1583


bernardnormier (approved):
It's not clear to me why we keep the msbuild/client and server folders. 

This is just temporary - they'll be eliminated in a follow-up PR?

externl (approved):

pepone:
> This is just temporary - they'll be eliminated in a follow-up PR?

Because we cannot have two C# project files in the same folder, we can rework these demos to be more similar to the ones from IceRPC  C# with Client/Server subdirectories

PR #206, opened 2024-05-20
Reformat C++ and add clang-format CI

externl:


.clang-format
```diff
@@ -0,0 +1,52 @@
...
+IndentPPDirectives: AfterHash
+KeepEmptyLinesAtTheStartOfBlocks: false
+NamespaceIndentation: All
+PackConstructorInitializers: CurrentLine
+PointerAlignment: Left
+SpaceAfterTemplateKeyword: false
+TabWidth: 4
+UseTab: Never
+---
+Language: ObjC
```
pepone, line 30: Do we have Objective-C code in the demos?

pepone (approved):

.clang-format
externl: No, good catch. I'll remove it

PR #214, opened 2024-06-03
Remove ice_timeout

externl:


pepone (approved):

csharp/Ice/callback/Client.cs
```diff
@@ -17,12 +17,12 @@ public static int Main(string[] args)
             // The new communicator is automatically destroyed (disposed) at the end of the
             // using statement
             //
-            using(var communicator = Ice.Util.initialize(ref args, "config.client"))
+            using (var communicator = Ice.Util.initialize(ref args, "config.client"))
```
bernardnormier, line 20: We should switch to using statements.

php/IceDiscovery/hello/Client.php
```diff
@@ -59,10 +59,6 @@
         {
             $p = $p->ice_secure(true);
         }
-        if(isset($_POST["timeout"]) and $_POST["timeout"] == "yes")
```
bernardnormier, line 62: I suspect this demo wants to demonstrate invocation timeouts.

bernardnormier (approved):

csharp/Ice/callback/Client.cs
externl: Sure, that's unrelated to this PR however.

PR #216, opened 2024-06-03
Add clang tidy and CI

externl:
Add .clang-tidy file with a few configured checks.


cpp/Ice/optional/ContactDBI.cpp
```diff
@@ -21,15 +21,15 @@ ContactDBI::addContact(
     contact->name = name;
     if (type)
     {
-        contact->type = std::move(type);
+        contact->type = type;
     }
     if (number)
     {
-        contact->number = std::move(number);
+        contact->number = number;
```
pepone, line 28: What was the problem with this `std::move` call?

pepone (approved):

cpp/Ice/optional/ContactDBI.cpp
externl: This type is trivially copyable. Using `std::move` doesn't do anything.

pepone: I guess we want to do `std::move(*number)` to move the string?

externl: Removing for number was a mistake. I've added it back.

PR #236, opened 2025-01-26
New Greeter c++ demo

bernardnormier:
This demo is a "port" of IceRPC C#'s Greeter example to Ice C++.

Note: it doesn't interoperate with the IceRPC example since the IceRPC example uses the `icerpc` protocol and the new Slice2 encoding.

This demo uses the synchronous API. I also want to add a greeterAsync demo in a follow-up PR that shows the async API.


make/Make.demos.rules
```diff
@@ -27,7 +27,7 @@ demo-programs           = client server publisher subscriber reader writer
 # The default demo sources for each demo executable.
 #
 demo-client-sources     = Client.$1 *.ice
-demo-server-sources     = Server.$1 *.ice *I.$1
+demo-server-sources     = Server.$1 *.ice *I.$1 Chatbot.$1
```
bernardnormier, line 30: I am not a fan of this code.

It should be easy for the user to figure out which files are compiled into which executable and currently it's all magic.

cpp/Ice/greeter/Server.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    ctrlCHandler.setCallback(
+        [communicator](int signal) // TODO: capturing the communicator by const reference is not a good model.
+        {
+            cout << "Caught signal " << signal << ", shutting down... " << flush;
+            communicator->shutdown(); // initiates the shutdown of the communicator
+        });
+
+    // Create an object adapter that listens for incoming requests and dispatches them to servants.
+    // TODO: should we include a -h in this endpoint?
```
pepone, line 29: Yes. I think that adding `-h 127.0.0.1` would be a bit clear.

pepone (approved):

cpp/Ice/greeter/Server.cpp
bernardnormier: The default shown by the demos should be to listen on all interfaces, which can be written as `tcp -p 4061` or `tcp -h \"::0\" -p 4061` or `tcp -h * -p 4061`.

make/Make.demos.rules
externl: I think ideally we would just have a simple Makefile per demo with a little more "duplication".

externl (approved):

PR #237, opened 2025-01-27
New greeterAsync c++ demo

bernardnormier:
This is the async variant of the Greeter demo.


cpp/Ice/greeterAsync/Client.cpp
```diff
@@ -0,0 +1,66 @@
...
+            }
+            promise.set_value();
+        });
+
+    // Wait for the response/exception callback to be called.
+    promise.get_future().get();
+
+    // Send a second request using the asynchronous future API provided by GreeterPrx.
+    future<string> greeting = greeter->greetAsync(name);
+    cout << greeting.get() << endl;
```
pepone, line 63: We should handle exceptions like above, otherwise this would crash when there the future throws.

pepone (approved):

cpp/Ice/greeterAsync/Chatbot.cpp
```diff
@@ -0,0 +1,43 @@
...
+        std::launch::async,
+        [name = std::move(name), response = std::move(response), exception = std::move(exception)]()
+        {
+            std::this_thread::sleep_for(std::chrono::seconds(2));
+            ostringstream os;
+            os << "Hello, " << name << "!";
+            response(os.str());
+        }));
+
+    // greetAsync completes immediately (releases the Ice server thread pool) while the task runs in the background.
```
externl, line 42:
```suggestion
    // greetAsync completes immediately (releasing the Ice server thread pool thread) while the task runs in the background.
```

cpp/Ice/greeterAsync/Chatbot.cpp
```diff
@@ -0,0 +1,43 @@
...
+    string name,
+    function<void(string_view)> response,
+    [[maybe_unused]] function<void(std::exception_ptr)> exception,
+    const Ice::Current&)
+{
+    cout << "Dispatching greet request { name = " << name << " }" << endl;
+
+    // Simulate a long-running background operation by using std::async.
+    // Note that we're moving all arguments except current into the lambda expression.
+    _tasks.push_back(std::async(
```
externl, line 32: It might be worth noting here that this is thread safe since there's only one configured thread pool thread (the default).

externl (approved):

cpp/Ice/greeterAsync/Client.cpp
bernardnormier: > this would crash when there the future throws.

Crashing is fine. That's what the plain greeter client does.

pepone: Then using set_exception in the callback example would be more consistent.

cpp/Ice/greeterAsync/Chatbot.cpp
bernardnormier: Fixed as suggested.

cpp/Ice/greeterAsync/Chatbot.cpp
bernardnormier: Fixed

cpp/Ice/greeterAsync/Client.cpp
bernardnormier: Fixed as suggested.

PR #241, opened 2025-01-27
New Greeter demo for C#

bernardnormier:
This new greeter demo is a port of the IceRPC slice/Greeter example. In particular, it uses a similar layout and build system.


csharp/Ice/Greeter/Server/Program.CancelKeyPressed.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) ZeroC, Inc.
+
+// Extends Program to provide the CancelKeyPressed property.
```
bernardnormier, line 3: In icerpc-csharp, this file is in a common folder.

csharp/Ice/Greeter/Server/Chatbot.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using VisitorCenter;
+
+namespace GreeterServer;
+
+/// <summary>A Chatbot is an Ice servant that implements Slice interface Greeter.</summary>
+internal class Chatbot : GreeterDisp_
+{
+    // Implements the abstract method greet from the GreeterDisp_ class generated by the Slice compiler.
+    public override string greet(string name, Ice.Current current)
```
bernardnormier, line 11: This is a "sync" dispatch, which is the simplest, and certainly ok when the dispatch is indeed sync.

Should I add a GreeterAsync demo that shows AMD, similar to the C++ greeterAsync demo?

csharp/Ice/Greeter/Client/Program.cs
```diff
@@ -0,0 +1,19 @@
...
+
+// GreeterPrx is a class generated by the Slice compiler. We create a proxy from a communicator and a "stringified
+// proxy" with the address of the target object.
+// If you run the server on a different computer, replace localhost in the string below with the server's hostname
+// or IP address.
+GreeterPrx greeter = GreeterPrxHelper.createProxy(communicator, "greeter:tcp -h localhost -p 4061");
+
+// Send a request to the remote object and get the response.
+// TODO: remap to GreetAsync with cs:identifier
+string greeting = await greeter.greetAsync(Environment.UserName);
```
bernardnormier, line 17: I think we should only demo the client-side async API. The sync API is for backwards compatibility - you probably shouldn't use it in new applications.

csharp/Ice/Greeter/README.md
```diff
@@ -0,0 +1,23 @@
+# Greeter
+
+The Greeter demo illustrates how to send a request and wait for the response.
+
+You can build the client and server applications with:
+
+``` shell
+dotnet build
```
bernardnormier, line 8: I was building with:
```
dotnet build /p:IceHome=$HOME/builds/ice /p:IceToolsPath=$HOME/builds/ice/cpp/bin
```

csharp/Ice/Greeter/Server/Chatbot.cs
externl: I wouldn't.

externl (approved):

pepone (approved):

PR #243, opened 2025-01-28
Add AMD server to Greeter demo in C#

bernardnormier:


csharp/Ice/Greeter/ServerAMD/Program.CancelKeyPressed.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) ZeroC, Inc.
```
bernardnormier, line 1: We should move this to a common shared directory in a follow-up PR.

pepone (approved):

externl (approved):

PR #248, opened 2025-01-30
Small fixes to Greeter demo

bernardnormier:


csharp/Ice/Greeter/ServerAMD/Chatbot.cs
```diff
@@ -9,7 +9,7 @@ internal class Chatbot : GreeterDisp_
 {
     // Implements the abstract method greet from the GreeterDisp_ class generated by the Slice compiler.
```
pepone, line 10: Should we use the actual name in the comment? there is no "greet" method in GreeDisp_

pepone (approved):

csharp/Ice/Greeter/ServerAMD/Chatbot.cs
bernardnormier: Yes, will fix.

PR #251, opened 2025-02-03
Update C++ Windows build to use Publish target from ice

pepone:
Companion to #3479 to use the new Publish target.


cpp/msbuild/ice.proj
```diff
@@ -17,62 +17,11 @@
         <Exec Command="powershell -ExecutionPolicy ByPass -Command &quot;[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12;(New-Object Net.WebClient).DownloadFile('$(NugetURL)', '$(NugetExe)')&quot;"/>
```
pepone, line 17: We can remove this file if we switch to CMake. is a helper project to build all the demos.

externl (approved):

PR #252, opened 2025-02-03
Initial CMake build support

externl:
This PR adds the initial support for building the C++ demos using CMake. The end goal is to use replace the Make and MSBuild based build systems with this.

Currently I've only added support for `greeter` and `greeterAsync`. 

We're using standard cmake setup which is different from our usual Make and MSBuild build systems. The compiled binaries will be available in the build directory (in a nested Release/Debug folder for multi config build systems).

Configure cmake:
```
cd Ice/greeter/
mkdir build
cd build
cmake ..
```

Build:
```
cmake --build .
```

Note: 

`Ice/greeter` uses a helper function I wrote (similar to one protobuf has)
`Ice/greeterAsync` does the slice compilation "by hand". 

I think having this function will be useful for more complicated builds. I'm undecided if we should use it in our demos. 

Note 2:

The build does not include lots of compiler build flags as it's not a best practice as they're often highly compiler dependent and are unnecessary for users. See https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-2.html

In the future we can add a profile for us to use for building with all of the warnings enabled. 

EDIT:

- Removed the top level CMakeLists.txt. It was just making things more painful.
- Always use the helper function


cpp/CMakeLists.txt
externl, line 1: We could consider getting rid of this and adding each dir to `cmake.sourceDirectory` as a list. The only reason I have it right now is for VS Code

cpp/Ice/greeter/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.16)
+
+project(IceGreeter CXX)
```
externl, line 3: This demo uses the `slice2cpp_generate` helper function.

cpp/Ice/greeter/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+include(../../cmake/common.cmake)
+
+add_executable(ice_greeter_client Client.cpp Greeter.ice)
+slice2cpp_generate(ice_greeter_client)
+set_target_properties(ice_greeter_client PROPERTIES OUTPUT_NAME client)
```
externl, line 9: Since we have a top level cmake file we need to have unique target names.

cpp/cmake/common.cmake
```diff
@@ -0,0 +1,95 @@
...
+set(CMAKE_CXX_STANDARD 20)
+else()
+set(CMAKE_CXX_STANDARD 17)
+endif()
+
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+find_package(Threads REQUIRED)
+
+if (NOT Ice_HOME)
```
externl, line 14: Yes, there is lots of cleanup and fixes to do here to support various settings and configurations.

cpp/cmake/common.cmake
```diff
@@ -0,0 +1,98 @@
...
+    if (DEFINED ENV{ICE_HOME})
+        set(Ice_HOME $ENV{ICE_HOME} )
+    else()
+        message(FATAL_ERROR "Ice_HOME not set")
```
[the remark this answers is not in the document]
externl: `Ice_HOME` is the `cmake` that a user should use to configure the Ice location. Getting from the env is just a convenience. The proper way is for the user to do something like `cmake .. -DIce_HOME=/path/to/ice`

cpp/cmake/common.cmake
```diff
@@ -0,0 +1,98 @@
...
+    endif()
+endif()
+
+if (NOT EXISTS ${Ice_HOME})
+    message(FATAL_ERROR "The specified Ice_HOME directory does not exist: ${Ice_HOME}")
+endif()
+
+find_program(Ice_SLICE2CPP_EXECUTABLE slice2cpp HINTS ${Ice_HOME}/cpp/bin PATH_SUFFIXES x64/Release x64/Debug)
```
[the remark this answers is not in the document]
externl: Yes, as I said in my comment

cpp/cmake/common.cmake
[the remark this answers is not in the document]
bernardnormier: The check above is specifically for the ICE_HOME environment variable.

cpp/Ice/greeterAsync/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+set(greeter_slice_header "${slice_output_dir}/Greeter.h")
+
+add_custom_command(
+    OUTPUT "${greeter_slice_source}" "${greeter_slice_header}"
+    COMMAND ${Ice_SLICE2CPP_EXECUTABLE} --output-dir "${slice_output_dir}" "${greeter_slice}"
+    DEPENDS "${greeter_slice}"
+    COMMENT "Compiling ${greeter_slice}"
+)
+
+add_executable(ice_greeter_async_client Client.cpp ${greeter_slice_source} ${greeter_slice_header})
```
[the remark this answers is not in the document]
externl: So they're dependencies.

cpp/Ice/greeterAsync/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+target_link_libraries(ice_greeter_async_client Ice::Ice)
+target_include_directories(ice_greeter_async_client PRIVATE ${slice_output_dir})
+set_target_properties(ice_greeter_async_client PROPERTIES OUTPUT_NAME client)
+
+add_executable(ice_greeter_async_server Server.cpp ChatBot.cpp ChatBot.h ${greeter_slice_source} ${greeter_slice_header})
```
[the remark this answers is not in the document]
externl: Ditto.

cpp/cmake/common.cmake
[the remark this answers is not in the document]
externl: The overall check is about `Ice_HOME`. If `Ice_HOME` is not set check ENV{ICE_HOME} else complain.

cpp/Ice/greeterAsync/CMakeLists.txt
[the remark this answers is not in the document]
pepone: Is it required? see https://discourse.cmake.org/t/c-header-files-and-dependency-management-in-cmake/10909

cpp/cmake/common.cmake
pepone: Did you try using FindIce CMake module?

https://github.com/Kitware/CMake/blob/master/Modules/FindIce.cmake

cpp/Ice/greeterAsync/CMakeLists.txt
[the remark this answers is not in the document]
externl: If you don't include them then they're not included in the list of headers in Xcode and Visual Studio projects. It's best to have them.

cpp/cmake/common.cmake
pepone: I ask because users will end up copying our `common.cmake` when it would be better for most users to use the FindIce.cmake module.

externl: Once we release hopefully we can switch to using `find_package(Ice)` but for now we'll have to use ours.

cpp/Ice/greeter/README.md
```diff
@@ -2,12 +2,37 @@
 
 The Greeter demo illustrates how to send a request and wait for the response.
 
+To build the demo run:
+
+```shell
+cmake -B build
+cmake --build build --config Release
```
bernardnormier, line 9: The default is Debug? Should we just show the default?

cpp/Ice/greeter/README.md
```diff
@@ -2,12 +2,37 @@
...
+
+```shell
+./build/server
 ```
-server
+
+**Windows:**
+
+```shell
+./build/Release/server
```
bernardnormier, line 23: Should this be:
```shell
build\Release\server
```
?

bernardnormier (approved):

cpp/Ice/greeter/README.md
externl: Yes.

externl: Note that this pathing is really generator specific and not specific. But since we're showing the defaults which are Make and "Visual Studio" I think it's fine to phrase like this.

cpp/Ice/greeter/README.md
```diff
@@ -2,12 +2,37 @@
...
+
+```shell
+./build/client
 ```
-client
+
+**Windows:**
+
+```shell
+./build/Release/client
```
pepone, line 37:
```suggestion
.\build\Release\client
```

To match server command above.

cpp/Ice/greeterAsync/README.md
```diff
@@ -2,12 +2,36 @@
...
+```shell
+./build/client
 ```
-client
-```
+
+**Windows:**
+
+```shell
+./build/Release/client
```
pepone, line 37:
```suggestion
.\build\Release\client
```

cpp/cmake/common.cmake
```diff
@@ -0,0 +1,92 @@
...
+    set(output_dir ${CMAKE_CURRENT_BINARY_DIR}/generated/${TARGET})
+    make_directory(${output_dir})
+
+    # Add the generated headers files to the target include directories
+    target_include_directories(${TARGET} PRIVATE ${output_dir})
+
+    # Process each Slice (.ice) file in the source list
+    # 1. Run the slice2cpp command to generate the header and source files
+    # 2. Add the generated files to the target sources
+    foreach(file IN LISTS sources)
```
pepone, line 73: This is fine for simple demos, but doesn't account for Slice dependencies.

pepone (approved):

cpp/cmake/common.cmake
externl: Yes

externl: We'll need to figure out how to handle this

PR #253, opened 2025-02-03
Add Greeter matlab demo

bernardnormier:


pepone (approved):

matlab/Ice/greeter/Greeter.ice
```diff
@@ -0,0 +1,15 @@
...
+#pragma once
+
+module VisitorCenter
+{
+    /// Represents a simple greeter.
+    interface Greeter
+    {
+        /// Creates a personalized greeting.
+        /// @param name: The name of the person to greet.
+        /// @returns: The greeting.
```
InsertCreativityHere, line 12:
```suggestion
        /// @param name The name of the person to greet.
        /// @return The greeting.
```
This doc-comment is incorrect:
1) There shouldn't be any colon separators in Slice1-Ice-Slice because Doxygen doesn't support it.
2) `@returns` isn't a supported tag. It should be `@return`.
The parser is letting it through because it isn't enforcing that there's a space after `@return`, a bug.
I'll add this to the comment-validation issue.

matlab/Ice/greeter/README.md
```diff
@@ -0,0 +1,12 @@
+# Greeter
+
+The Greeter demo illustrates how to write a client application with Ice for MATLAB.
+
+Ice for MATLAB supports only client-side applications. As a result, you need to first start a Greeter server implemented
+in a language with server-side support, such as Python, Java or C#.
```
InsertCreativityHere, line 6: Grammar pedantry:
```suggestion
Ice for MATLAB supports only client-side applications. As a result, you first need to start a Greeter server implemented
in a language with server-side support, such as Python, Java, or C#.
```

InsertCreativityHere (approved):
Looks good to me, just some random small things.

PR #257, opened 2025-02-04
Add CMakeLists.txt for cpp/Ice demos

externl:
This PR adds CMakeLists.txt to the `cpp/Ice` demos that we plan on keeping.


bernardnormier (approved):

cpp/Ice/invoke/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+cmake_minimum_required(VERSION 3.16)
+
+project(greeter CXX)
```
InsertCreativityHere, line 3:
```suggestion
project(invoke CXX)
```

InsertCreativityHere (approved):
I didn't look at the README's because it looks like they're just cookie-cutters of each other.

But I checked all the other files, and everything looks fine to me!

PR #259, opened 2025-02-04
Php greeter

bernardnormier:
We may want to add a second HTML version of this demo. Or maybe make "secure" HTML.


php/Ice/greeter/Client.php
```diff
@@ -0,0 +1,21 @@
...
+// GreeterPrxHelper is a class generated by the Slice compiler. We create a Greeter proxy from a communicator and a
+// "stringified proxy" with the address of the target object.
+// If you run the server on a different computer, replace localhost in the string below with the server's hostname
+// or IP address.
+$greeter = VisitorCenter\GreeterPrxHelper::createProxy($communicator, 'greeter:tcp -h localhost -p 4061');
+
+// Send a request to the remote object and get the response.
+$greeting = $greeter->greet(get_current_user());
+
+echo "$greeting\n";
```
bernardnormier, line 20: I assume that's the correct syntax. Please check.

php/Ice/greeter/Client.php
```diff
@@ -0,0 +1,21 @@
+<?php
+// Copyright (c) ZeroC, Inc.
+
+require_once 'Ice.php';
+require_once 'Greeter.php';
+
+// Create an Ice communicator to initialize the Ice runtime. This communicator is destroyed automatically at the end of
```
bernardnormier, line 7: I verified the communicator is indeed destroyed even though there is no call to ->destroy

externl (approved):

PR #260, opened 2025-02-05
Use CtrlCHandler in IceStorm/clock (Swift)

### LINKED ISSUE #75: Improve termination of Swift servants in response to signals
The pattern used in IceStorm/clock subscriber, isn't simple and also is not thread safe, there is a chance for exit call run when GDC global queue is still releasing the communicator reference in the signal event handler

https://github.com/zeroc-ice/ice-demos/blob/d3d8881d3e2351a6a196be29c16cd7aaaa4e63a0/swift/IceStorm/clock/Sources/Subscriber/main.swift#L41-L48

For the publisher we do all the work in the main thread including the signal handling, but it is still not straightforward and will be good if we come up with a better solution

https://github.com/zeroc-ice/ice-demos/blob/d3d8881d3e2351a6a196be29c16cd7aaaa4e63a0/swift/IceStorm/clock/Sources/Publisher/main.swift#L35-L45

bernardnormier:
This PR updates the IceStorm/clock demo to use async/await and the CtrlCHandler.

The focus of this PR is fixing #75. This demo should be reviewed / improved in a follow-up PR.

Fixes #75.


swift/IceStorm/clock/slice/Clock.ice
bernardnormier, line 1: I used Greeter as a model. Could we put this slice directory in Sources? I would find it clearer, since the Slice file is a source file.

swift/IceStorm/clock/Sources/Publisher/main.swift
```diff
@@ -113,27 +98,34 @@ func run() -> Int32 {
...
+        return 0
     } catch {
         print("Error: \(error)\n")
         return 1
     }
 }
 
-exit(run())
+let ctrlCHandler = Ice.CtrlCHandler()
+exit(await run(ctrlCHandler: ctrlCHandler))
```
bernardnormier, line 131: I am not fond of this "exit(run)" pattern, we should replace it.

With this pattern, it's critical to create the CtrlCHandler before the first await call.

swift/IceStorm/clock/Sources/Publisher/main.swift
```diff
@@ -113,27 +98,34 @@ func run() -> Int32 {
...
-                print("Error: \(error)\n")
-                communicator.destroy()
-                exit(1)
+        // Send a tick every second until cancelled
+        let task = Task {
+            while true {
+                do {
+                    try await clock.tick(dateFormatter.string(from: Date()))
+                } catch let error as Ice.LocalException {
+                    print("tick invocation failed with: \(error.message)")
```
bernardnormier, line 107: The plain \(error) for a local exception displays the full C++ stack trace. We should double-check that's the desired behavior.

swift/IceStorm/clock/Sources/Publisher/main.swift
externl: It's also unnecessary

swift/IceStorm/clock/Sources/Publisher/main.swift
externl: I think it's undesirable.

swift/IceStorm/clock/slice/Clock.ice
externl: Yes, makes sense to me.

pepone (approved):

PR #261, opened 2025-02-05
Cleanup IceStorm/clock demo

bernardnormier:
This PR cleans up the IceStorm/clock demo in Swift.

The logic remains the same.  3 main changes:
 - eliminate run function
 - the OA Endpoints are now either all interfaces (no -h) or -h 127.0.0.1, never a hostname
 - more comments


swift/IceStorm/clock/README.md
```diff
@@ -1,168 +1,131 @@
...
-This configuration assumes there is a subdirectory named db in the
-current working directory.
+This configuration assumes there is a subdirectory named db in the current working directory.
 
 In a separate window:
 
-```
-./Build/subscriber
+```shell
+swift run Subscriber
```
bernardnormier, line 18: Joe wants to switch to lowercase subscriber/publisher/client/server - for a follow-up PR.

swift/IceStorm/clock/Sources/Clock.ice
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ZeroC, Inc.
```
bernardnormier, line 1: Moved to Sources/

swift/IceStorm/clock/Sources/Clock.ice
```diff
@@ -0,0 +1,14 @@
...
+
+#pragma once
+
+module Demo
+{
+    /// The interface implemented by the subscriber. It accepts time updates.
```
bernardnormier, line 7: Note the best interface name, but I didn't want to change it.

swift/IceStorm/clock/Sources/Subscriber/main.swift
```diff
@@ -8,12 +8,15 @@ import IceStorm
...
 func usage() {
-    print("Usage: [--batch] [--datagram|--twoway|--ordered|--oneway] [--retryCount count] [--id id] [topic]")
+    print(
+        "Usage: [--batch] [--datagram|--twoway|--ordered|--oneway] [--retryCount count] [--id id] [topic]"
+    )
 }
 
+/// ClockI is an Ice servant that implements Slice interface Clock.
 class ClockI: Clock {
-    func tick(time date: String, current _: Ice.Current) throws {
```
bernardnormier, line 15: Unclear why we "renamed" time to date.

swift/IceStorm/clock/Sources/Subscriber/main.swift
```diff
@@ -25,150 +28,136 @@ enum Option: String {
...
-                break
-            }
-
-            if oldoption != option, oldoption != .none {
-                usage()
-                return 1
-            }
-        }
+// Create an Ice communicator to initialize the Ice runtime.
+var args = [String](CommandLine.arguments.dropFirst())
```
bernardnormier, line 36: We should show the same in the Greeter demo and hopefully explain this dropFirst.

swift/IceStorm/clock/README.md
```diff
@@ -1,168 +1,131 @@
...
+```shell
+swift run Subscriber --id <id>
 ```
 
-This option specifies a unique identity for this subscriber. When
-you use this option, you should also run the subscriber on a fixed
-port by setting the `Clock.Subscriber.Endpoints` property. For
-example:
+This option specifies a unique identity for this subscriber. When you use this option, you should also run the
+subscriber on a fixed port by setting the `Clock.Subscriber.Endpoints` property. For example:
```
bernardnormier, line 74: That's why we parse with the Clock prefix in the Subscriber. That's code for advanced users!

externl (approved):

swift/IceStorm/clock/Sources/Publisher/main.swift
```diff
@@ -18,114 +18,106 @@ enum Option: String {
...
-        } catch is IceStorm.NoSuchTopic {
-            do {
-                topic = try await manager.create(topicName)
-            } catch is IceStorm.TopicExists {
-                print("temporary error. try again.")
-                return 1
-            }
-        }
+for var i in 0..<args.count {
+    let oldOption = option
```
pepone, line 38: Can't we get rid of `oldOption` and  also remove 

```
if oldOption != option, oldOption != .none {
        usage()
        exit(1)
    }
```

We can check this before doing `option = o`

swift/IceStorm/clock/Sources/Publisher/main.swift
```diff
@@ -18,114 +18,106 @@ enum Option: String {
...
-        }
+    if oldOption != option, oldOption != .none {
+        usage()
+        exit(1)
+    }
+}
 
-        let clock = uncheckedCast(prx: publisher, type: ClockPrx.self)
+// Create the topic manager proxy.
+let manager = try uncheckedCast(
```
pepone, line 59: Isn't there a generic propertyToProxy for Swift?

pepone (approved):
Looks good. The demo has too many arguments which makes it complicated, there is too much code for parsing arguments in the subscriber.  I think we should aim for a simple IceStorm demo.

swift/IceStorm/clock/Sources/Publisher/main.swift
bernardnormier: There is currently only:
```
Ice/Communicator.swift
61:    func propertyToProxy(_ property: String) throws -> ObjectPrx?
```

swift/IceStorm/clock/Sources/Publisher/main.swift
bernardnormier: I didn't change this code (except oldoption => oldOption) ... a lot more demo code to cleanup.

PR #262, opened 2025-02-05
Add cpp/Chat cmake files

externl:
This PR adds CMakeLists.txt for the chat demos (client and server).

It also reworks the cmake files a use a standard custom package config.


cpp/cmake/IceConfig.cmake
```diff
@@ -0,0 +1,116 @@
...
+  else()
+    message(FATAL_ERROR "Ice_HOME not set")
+  endif()
+endif()
+
+if (NOT EXISTS ${Ice_HOME})
+  message(FATAL_ERROR "The specified Ice_HOME directory does not exist: ${Ice_HOME}")
+endif()
+
+find_program(Ice_SLICE2CPP_EXECUTABLE slice2cpp HINTS ${Ice_HOME}/cpp/bin PATH_SUFFIXES x64/Release x64/Debug)
```
pepone, line 13: Should we add Win32/Release Win32/Debug

pepone (approved):

cpp/cmake/IceConfig.cmake
externl: Win32/x64 with cmake is generally figured out from the target os. See https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html

I know we support Win32 builds with Ice, but I don't really see it being very useful and having to have some extra config to handle this.

pepone: You can still want to use Win32 binaries in x64 platforms, we assume nobody runs a Win32 native platform this days. Same for our x86 linux builds. Anyways this is something we can figure out later.

PR #266, opened 2025-02-06
Add IceBox CMake files

externl:
This PR adds CMake support to the IceBox/hello demo. 

Due to the nature of CMake's support of multi-config generators there are two places the HelloService library may end up by default: `./build/` and `./build/Release`. This causes a bit of havoc with the config file locations.

Having 4 config files (most just full of superfluous config options) for this "simple" demo seems like a lot. 

- This was the only demo using an `XXXadmin` tool to shutdown a service. We don't do this for the IceGrid demos, it's unnecessary here.
- I've removed SSL configuration as well as the "secure" toggle code in the client

IMO these changes make the demo much easier to understand and digest.


cpp/IceBox/hello/Client.cpp
```diff
@@ -53,20 +53,13 @@ void menu();
...
 run(const shared_ptr<Ice::Communicator>& communicator)
 {
-    auto twoway =
-        Ice::checkedCast<HelloPrx>(communicator->propertyToProxy("Hello.Proxy")->ice_twoway()->ice_secure(false));
-    if (!twoway)
-    {
-        cerr << "invalid proxy" << endl;
-        return 1;
-    }
+    HelloPrx twoway{communicator, "hello:tcp -h localhost -p 10000:udp -h localhost -p 10000"};
```
pepone, line 56: We should further simplify in a follow up PR and have IceBox/greeter, inline with `Ice/greeter`.

cpp/IceBox/hello/README.md
```diff
@@ -1,23 +1,42 @@
...
+start the IceBox server from the directory containing the `HelloService` library:
+
+**Linux/macOS:**
+
+```shell
+icebox --IceBox.Service.Hello="./build/HelloService:create"
 ```
-icebox --Ice.Config=config.icebox
+
+**Linux/macOS:**
```
pepone, line 19: Was this supposed to be Windows?

cpp/IceBox/hello/README.md
```diff
@@ -1,23 +1,42 @@
...
+
+**Windows:**
+
+```shell
+build\Release\client
 ```
 
-Linux 32-bit
-------------
+## Linux 32-bit
```
pepone, line 39: We need to add support for this builds in CMake.

cpp/IceBox/hello/Client.cpp
externl: and remove all udp support?

cpp/IceBox/hello/HelloServiceI.cpp
```diff
@@ -20,10 +22,10 @@ HelloServiceI::start(
     const shared_ptr<Ice::Communicator>& communicator,
     const Ice::StringSeq& /*args*/)
 {
-    _adapter = communicator->createObjectAdapter(name);
-    auto hello = make_shared<HelloI>();
-    _adapter->add(hello, Ice::stringToIdentity("hello"));
+    _adapter = communicator->createObjectAdapterWithEndpoints("Hello", "tcp -p 10000:udp -p 10000");
```
bernardnormier, line 25: (For a separate PR): we should stop using udp with random demos and showing off multi-endpoint OAs.

cpp/IceBox/hello/HelloServiceI.cpp
```diff
@@ -1,8 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
-#include <HelloI.h>
-#include <HelloServiceI.h>
+#include "HelloServiceI.h"
+#include "HelloI.h"
+
 #include <Ice/Ice.h>
+#include <iostream>
```
bernardnormier, line 7: I suggest adding a blank like between Ice/IceStorm etc. headers and other system headers.

cpp/IceBox/hello/HelloServiceI.cpp
```diff
@@ -20,10 +22,10 @@ HelloServiceI::start(
     const shared_ptr<Ice::Communicator>& communicator,
     const Ice::StringSeq& /*args*/)
 {
```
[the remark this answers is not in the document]
externl: This is not related to my PR

cpp/IceBox/hello/README.md
```diff
@@ -1,23 +1,48 @@
...
+
+```shell
+DYLD_LIBRARY_PATH=./build icebox --IceBox.Service.Hello="HelloService:create --Ice.Trace.Dispatch"
 ```
-client
+
+**Windows:**
+
+```shell
+set PATH=%PATH%;./build/Release
```
bernardnormier, line 27: I would not use '/' in a Windows PATH. Does it actually work?

bernardnormier (approved):

PR #268, opened 2025-02-06
Add IceDiscovery CMake files

externl:


cpp/IceDiscovery/hello/README.md
```diff
@@ -1,15 +1,40 @@
+# IceGrid Hello
```
pepone, line 1: IceDiscovery?

pepone (approved):

PR #272, opened 2025-02-06
Add Secure demo in C#

bernardnormier:


csharp/Ice/Secure/Client/Program.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) ZeroC, Inc.
+
+// Slice module VisitorCenter in Greeter.ice maps to C# namespace VisitorCenter.
+using VisitorCenter;
+
+// Create an Ice communicator to initialize the Ice runtime. The communicator gets its properties from file
+// "config.client".
+using Ice.Communicator communicator = Ice.Util.initialize(ref args, "config.client");
```
pepone, line 8: Should we instead show case the new APIs using `SslClientAuthenticationOptions`?

IceSSL properties are mostly for compatibility, but the native APIs should be preferred.

externl (approved):

bernardnormier:
> Looks good, but I think this demo should use the API instead. And use the `SslClientAuthenticationOptions` and `SslServerAuthenticationOptions` to configure SSL.

I agree. Please go ahead and fix it.

PR #275, opened 2025-02-07
Rework filesystem demo in C++

bernardnormier:
This PR moves cpp/Manual/simpleFilesystem to cpp/Ice/filesystem and reworks the demo.


cpp/Ice/filesystem/Client.cpp
```diff
@@ -0,0 +1,57 @@
...
+listRecursive(const DirectoryPrx& dir, size_t depth = 0)
+{
+    const string indent(++depth, '\t');
+
+    // The node proxies returned by list() are never null.
+    NodeSeq contents = dir->list();
+
+    for (const optional<NodePrx>& node : contents)
+    {
+        auto subdir = Ice::checkedCast<DirectoryPrx>(node);
```
pepone, line 24: I wonder if we should just use different categories for the node types, instead of using checked casts.

cpp/Manual/simpleFilesystem/FilesystemI.h
```diff
@@ -1,60 +0,0 @@
...
-
-#ifndef FilesystemI_h
-#define FilesystemI_h
-
-#include <Filesystem.h>
-#include <Ice/Ice.h>
-
-#ifdef _MSC_VER
-#    pragma warning(push)
-#    pragma warning(disable : 4250) // ... : inherits ... via dominance
```
bernardnormier, line 11: We may need to add this back on Windows.

pepone (approved):

cpp/Ice/filesystem/Server.cpp
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) ZeroC, Inc.
+
+#include "Memory.h"
+
+#include <Ice/Ice.h>
```
bernardnormier, line 5: I changed my mind on keep Ice.h and other system includes together.

cpp/Ice/filesystem/Memory.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+void
+MNode::activate(const Ice::ObjectAdapterPtr& adapter)
+{
+    auto proxy = adapter->add<Filesystem::NodePrx>(shared_from_this(), _id);
+    if (_parent)
+    {
+        _parent->addChild(proxy);
+    }
+}
```
externl, line 32: _parent is only used for the name and and adding a child. I think we should just do all of this in the server setup. 

1. activate doesn't "activate" anything.
2. `if(_parent)` is true 99% of the time
3. we could remove our dependance on shared_from_this.

externl: It's only 3 extra lines of code. For the 3 things we call activate on that aren't the root node.

bernardnormier: Simplified as suggested.

externl (approved):

PR #277, opened 2025-02-07
Rename secure -> config and change its focus (C++)

bernardnormier:
This PR renames the Ice/secure demo to Ice/config and changes its focus to be configuration properties and config files.

In particular, it provides a Greeter server we'll be able to use from a JS client.


cpp/Ice/config/README.md
externl, line 1: I think it would be better to show:

```
./build/client --Ice.Config=config.client
./build/client --Greeter.Proxy="default -h localhost -p 4061"
```

and not hardcode "config.client" in Client.cpp

pepone (approved):

cpp/Ice/config/README.md
bernardnormier: I find it clearer with config.client/config.server hardcoded in the programs.

externl (approved):

PR #278, opened 2025-02-07
Improvements to IceConfig.cmake

externl:
- Lots of improvements IceConfig.cake
- Fix workflow to not used `find -exec` as failure from the exec command do not cause find to return with an error.

### DIFF
--- a/cpp/cmake/IceConfig.cmake
+++ b/cpp/cmake/IceConfig.cmake
@@ -1,3 +1,4 @@
+
 if (NOT Ice_HOME)
   if (DEFINED ENV{ICE_HOME})
     set(Ice_HOME $ENV{ICE_HOME} CACHE PATH "Path to the Ice installation directory")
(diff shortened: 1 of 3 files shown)


cpp/cmake/IceConfig.cmake
```diff
@@ -10,69 +11,173 @@ if (NOT EXISTS ${Ice_HOME})
...
 
-find_program(Ice_SLICE2CPP_EXECUTABLE slice2cpp HINTS ${Ice_HOME}/cpp/bin PATH_SUFFIXES x64/Release x64/Debug)
+# List of all available Ice components
+set(Ice_AVAILABLE_COMPONENTS DataStorm Glacier2 Ice IceBox IceGrid IceStorm)
 
-if(NOT DEFINED Ice_SLICE_DIR AND EXISTS ${Ice_HOME}/slice)
-  set(Ice_SLICE_DIR ${Ice_HOME}/slice CACHE PATH "Path to the Ice Slice files directory")
+# IceBT is only available on Linux
+# TODO: Check for required dependency
+if(UNIX AND NOT APPLE)
```
pepone, line 19: We don't always build IceBT on Linux, would be good if we can check if the file is present.

pepone (approved):
Looks good

PR #280, opened 2025-02-09
New Invoke demo in C#

bernardnormier:
This PR implements a new Invoke demo in C# (Invoke2 for now, until we retire invoke).

I initially wanted to call it "SliceFree" and provide a Slice-free demo. However, even though this demo does not provide any Slice definition or use any generated code, it still uses InputStream / OutputStream to marshal/unmarshal data (strings) using the Ice encoding.

Using InputStream / OutputStream is unavoidable with Ice, even if we just want to transfer bytes.

This demo is primarily about demonstrating `ice_invokeAsync` (on Ice.ObjectPrx) and `dispatchAsync` (on Ice.Object).
In practice, these methods are more useful to build a Forwarding servant than to send Slice requests "by hand". Maybe we should drop this Invoke demo and provide a Forwarder demo instead?

### DIFF
--- a/csharp/Ice/Invoke2/Client/Client.csproj
+++ b/csharp/Ice/Invoke2/Client/Client.csproj
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <OutputType>Exe</OutputType>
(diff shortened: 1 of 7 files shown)


pepone:
>  In practice, these methods are more useful to build a Forwarding servant than to send Slice requests "by hand". Maybe we should drop this Invoke demo and provide a Forwarder demo instead?

Yes. I think a Forwarder demo makes more sense. Create Ice requests using the OutpuStream API is not a common use case.

PR #282, opened 2025-02-10
Add IceBT CMakeLists.txt

externl:


.github/workflows/ci.yml
externl, line 1: I thinks preferable that the IceBT demo build fails if IceBT is missing. So we just filter it from CI.

cpp/cmake/IceConfig.cmake
```diff
@@ -11,27 +11,6 @@ if (NOT EXISTS ${Ice_HOME})
   message(FATAL_ERROR "The specified Ice_HOME directory does not exist: ${Ice_HOME}")
 endif()
 
-# List of all available Ice components
-set(Ice_AVAILABLE_COMPONENTS DataStorm Glacier2 Ice IceBox IceGrid IceStorm)
```
externl, line 15: The `_AVAILABLE_COMPONENTS` was unnecessary

PR #284, opened 2025-02-10
Fix namespace in demo servers (C++ and C#)

bernardnormier:
This demo updates the new demos (Greeter, Config) to use namespace Server for the server implementation.

It's consistent with the C# naming conventions and works fine with Ice since Ice doesn't define a Server class.


cpp/Ice/config/Chatbot.cpp
```diff
@@ -6,10 +6,9 @@
 #include <sstream>
 
 using namespace std;
-using namespace ConfigServer;
```
bernardnormier, line 9: Also removed `using namespace` to make it clearer what comes from the Server namespace.

csharp/Ice/Greeter/Server/Program.cs
```diff
@@ -1,15 +1,13 @@
 // Copyright (c) ZeroC, Inc.
 
-using GreeterServer;
```
bernardnormier, line 3: Same as in C++: it's clearer without this using.

swift/Ice/greeter/Sources/Server/Chatbot.swift
```diff
@@ -3,7 +3,7 @@
 import Ice
 
 /// A Chatbot is an Ice servant that implements Slice interface Greeter.
-class ChatBot: Greeter {
+class Chatbot: Greeter {
```
bernardnormier, line 6: Unrelated fix.

pepone (approved):

PR #285, opened 2025-02-11
New forwarder demo in C++

bernardnormier:
This is the C++ counterpart of #283.


cpp/Ice/forwarder/Forwarder.cpp
```diff
@@ -0,0 +1,51 @@
...
+    // Read the encapsulation containing the in-parameters.
+    const std::byte* inEncapsulationStart = nullptr;
+    int32_t inEncapsulationSize = 0;
+    request.inputStream().readEncapsulation(inEncapsulationStart, inEncapsulationSize);
+
+    // Make the invocation asynchronously. This call reports most exceptions through its exception callback.
+    target.ice_invokeAsync(
+        current.operation,
+        current.mode,
+        make_pair(inEncapsulationStart, inEncapsulationStart + inEncapsulationSize),
```
pepone, line 33: This is safe because the encapsulation for the input parameters is synchronously copied to OutgoingAsync. We don’t have a way to avoid this copy, right?

bernardnormier: Yes, the "marshaling" is always synchronous for Ice invocations, including ice_invokeAsync.

And indeed, the marshaling - even of bytes - always allocates a new buffer. There is no public API to transfer the ownership of the in-params InputStream to the in-params OutputStream.

For colloc, we have an _internal_ API that does the opposite: "converts" an OutputStream into an InputStream. 
https://github.com/zeroc-ice/ice/blob/4f7ba6a20a54c43c54891fc74494843107d78cd4/cpp/src/Ice/CollocatedRequestHandler.cpp#L136

https://github.com/zeroc-ice/ice/blob/4f7ba6a20a54c43c54891fc74494843107d78cd4/cpp/include/Ice/InputStream.h#L103

Note that adopt is false, most likely for retries.

cpp/Ice/forwarder/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+add_executable(client Client.cpp Greeter.ice)
+slice2cpp_generate(client)
+target_link_libraries(client Ice::Ice)
+
+add_executable(server Server.cpp Chatbot.cpp Chatbot.h Greeter.ice)
+slice2cpp_generate(server)
+target_link_libraries(server Ice::Ice)
+
+add_executable(forwardingserver Forwarder.cpp Forwarder.h ForwardingServer.cpp)
```
externl, line 15: I would just call it `forwarder`

externl (approved):

cpp/Ice/forwarder/CMakeLists.txt
bernardnormier: forwarding server is more consistent with the file names and C# names.

csharp/Ice/Forwarder/ForwardingServer/Forwarder.cs
```diff
@@ -14,7 +14,7 @@ internal class Forwarder : Ice.Object
     // Implements abstract method dispatchAsync defined on Ice.Object.
     public async ValueTask<OutgoingResponse> dispatchAsync(IncomingRequest request)
     {
-        // Change the identity and facet of _target to match the identity and facet in the incoming request.
+        // Create a proxy with the the desired identity and facet.
```
InsertCreativityHere, line 17: Double 'the' here.
```suggestion
        // Create a proxy with the desired identity and facet.
```

csharp/Ice/Forwarder/ForwardingServer/Forwarder.cs
```diff
@@ -23,8 +23,8 @@ public async ValueTask<OutgoingResponse> dispatchAsync(IncomingRequest request)
             target = target.ice_oneway();
         }
 
-        // Make the invocation asynchronously. If `ice_invokeAsync` throws an Ice.LocalException that cannot be
-        // marshaled (such as Ice.ConnectionRefusedException), the object adapter that calls `dispatchAsync` converts it
```
InsertCreativityHere, line 27: Why remove these?
I think using quotes to distinguish between **_english_** and **_code_** is helpful to comprehension, and wish that we did more of this. Although I would use single quotes instead of backticks for these (personally).

InsertCreativityHere (approved):
Looks good.
I built and tested it out on Windows, everything worked fine, and was very clear IMO.

PR #286, opened 2025-02-11
Delete Chat, latency, throughput, interleaved, nested and more C++ demos

bernardnormier:
This PR deletes a number of C++ demos.

Some of these deletes demos show features we don't expect users to implement (locator, plugin, invoke). Others are too complicated (Chat, nested).  The benchmarks (latency, throughput, interleaved) shouldn't be demos. 

Finally, Manual/printer is replaced by Ice/greeter.


externl (approved):

InsertCreativityHere:
`cpp/README.md` needs some updates after this.
It still talks about Manual and Chat and probably other unnecessary things now.

pepone (approved):

InsertCreativityHere (approved):
Apart from the READMEs, obviously everything else looks good.

PR #288, opened 2025-02-11
Update filesystem demo in C#

bernardnormier:
It's now in sync with the C++ version.


cpp/Ice/filesystem/Memory.cpp
```diff
@@ -3,16 +3,15 @@
 #include "Memory.h"
 
 using namespace std;
-using namespace Memory;
```
bernardnormier, line 6: Switch to namespace Server and no using namespace like in other C++ demos.

csharp/Ice/Filesystem/Server/MDirectory.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Server;
+
+/// <summary>Provides an in-memory implementation of the Slice interface Directory.</summary>
+internal class MDirectory : Filesystem.DirectoryDisp_
+{
+    private readonly string _name;
+    private readonly List<Filesystem.NodePrx> _contents = new();
+
+    /// <inheritdoc/>
```
bernardnormier, line 11: There is currently nothing to inherit currently as slice2cs neglects to generate doc-comments for server-side interfaces / classes.

csharp/Ice/Filesystem/Server/MDirectory.cs
```diff
@@ -0,0 +1,24 @@
...
+    public override string Name(Ice.Current current) => _name;
```
bernardnormier, line 12: Unlike the C++ version, we can't define a MNode and reuse it. That's because XxxDisp_ is a class.

externl (approved):

InsertCreativityHere (approved):

PR #289, opened 2025-02-11
Add new callback demo in C#

bernardnormier:
This PR implements a new callback demo in C#. It replaces the existing callback demo.

The callback in this demo is `AlarmClock::ring`, which is more interesting than a void-returning operation. I plan to reuse the same Slice definitions and most of the code for a new bidir demo.

Typical output (updated):
```text
Client % dotnet run
Listening on ephemeral port...
Wake-up call scheduled, falling asleep...
Dispatching ring request { message = 'It's time to wake up!' }
Returning Snooze to request more time.
Dispatching ring request { message = 'No more snoozing!' }
Returning Stop to stop the alarm.
Stop button pressed, exiting...
```

```text
Server % dotnet run
Listening on port 4061...
Dispatching wakeMeUp request { alarmClock = 'alarmClock -t -e 1.1:tcp -h studio.local -p 56136 -t 60000', timeStamp = '2/12/2025 9:42:04 AM' }
^CCaught Ctrl+C, exiting...
```


csharp/Ice/Callback/Client/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+using EarlyRiser;
+
+// Create an Ice communicator to initialize the Ice runtime.
+using Ice.Communicator communicator = Ice.Util.initialize(ref args);
+
+// Create an object adapter that listens for incoming requests and dispatches them to servants.
+Ice.ObjectAdapter adapter = communicator.createObjectAdapterWithEndpoints("AlarmClockAdapter", "tcp -p 10000");
+
+// Register the MockAlarmClock servant with the adapter, and get an alarm clock proxy.
+// The identity of this Ice object is a UUID, which is suitable for transient objects.
```
bernardnormier, line 13: On second thought, it would better to use a fixed identity like alarmClock. We no longer advocate for globally unique identities.

csharp/Ice/Callback/slice/AlarmClock.ice
```diff
@@ -0,0 +1,34 @@
...
+        ButtonPressed ring(string message);
+    }
+
+    /// Represents the wake up service provided by the server.
+    interface WakeUpService
+    {
+        /// Schedules a call to {@link AlarmClock::ring} at the specified time.
+        /// @param alarmClock A proxy to the alarm clock to ring.
+        /// @param timeStamp The time to ring the alarm clock. It's encoded as the number of ticks (100 nanoseconds)
+        /// since January 1, 0001 00:00:00 UTC in the Gregorian calendar.
```
bernardnormier, line 30: See https://github.com/icerpc/icerpc-slice/blob/main/WellKnownTypes/TimeStamp.slice

csharp/Ice/Callback/Server/SimpleWakeUpService.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        var timeStampDateTime = new DateTime(timeStamp, DateTimeKind.Utc);
+
+        Console.WriteLine(
+            $"Dispatching wakeMeUp request {{ alarmClock = '{alarmClock}', timeStamp = '{timeStampDateTime.ToLocalTime()}' }}");
+
+        // Start a background task to ring the alarm clock.
+        _ = Task.Run(async () =>
+        {
+            // Wait until the specified time.
```
pepone, line 28: Should we handle exceptions here, event if it is just to ignore them? We don't want an unobserved task exception to terminate the server process.

pepone (approved):
Looks good.

externl (approved):
Looks good!

csharp/Ice/Callback/Server/SimpleWakeUpService.cs
pepone: I thought the default was to terminate the process.

> This event provides a mechanism to prevent exception escalation policy (which, by default, terminates the process) from triggering.

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=net-9.0

csharp/Ice/Callback/Client/Program.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) ZeroC, Inc.
+
+// Slice module EarlyRiser in AlarmClock.ice maps to C# namespace EarlyRiser.
+using EarlyRiser;
+
+// Create an Ice communicator to initialize the Ice runtime.
+using Ice.Communicator communicator = Ice.Util.initialize(ref args);
+
+// Create an object adapter that listens for incoming requests and dispatches them to servants.
+Ice.ObjectAdapter adapter = communicator.createObjectAdapterWithEndpoints("AlarmClockAdapter", "tcp -p 10000");
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

csharp/Ice/Callback/Server/SimpleWakeUpService.cs
bernardnormier: See the Remarks section on that page. Or give it a try with this demo.

PR #290, opened 2025-02-12
Fix ChatbotAMD to clear completed futures

bernardnormier:
The initial version was not quite correct as _tasks would grow forever.


cpp/Ice/greeter/ChatbotAMD.cpp
```diff
@@ -43,6 +45,15 @@ ServerAMD::Chatbot::greetAsync(
             response(os.str());
         }));
 
+    // We don't want the _tasks vector to grow forever so remove all completed task here, without waiting.
+    // TODO: switch to std::erase_if when we can use C++20.
```
bernardnormier, line 49: The pre-C++20 syntax is unfortunately not very readable. See item 32 in Scott Meyers' Effective STL.

pepone (approved):

externl (approved):

PR #291, opened 2025-02-12
New Bidir demo for C#

bernardnormier:
This PR replaces the bidir demo by a new demo derived from Callback.

PR #295, opened 2025-02-12
New Ice/bidir demo for C++

bernardnormier:
This is the C++ counterpart of #291.


externl (approved):

cpp/Ice/bidir/BidirWakeUpService.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // Create a proxy to the client's alarm clock. This connection-bound proxy is called a "fixed proxy".
+    auto alarmClock = connection->createProxy<AlarmClockPrx>(Ice::stringToIdentity("alarmClock"));
+
+    // Schedule a wake-up call in a background task.
+    _tasks.emplace_back(std::async(
+        std::launch::async,
+        [alarmClock = std::move(alarmClock), timePoint]()
+        {
+            // Sleep until the specified time point.
```
pepone, line 55: I guess is ok for simples demos to no handle exceptions in this remote requests.

pepone (approved):

cpp/Ice/bidir/BidirWakeUpService.cpp
bernardnormier: Yes, the exception handling is not what this demo is about. Less extra code = clearer demo.

Also, note that an exception does not result in a crash here. We only wait the futures (tasks), never get their results.

PR #297, opened 2025-02-13
New Middleware demo in C#

bernardnormier:
This PR adds a new middleware demo in C#. It will replace the interceptor demo.


csharp/Ice/Middleware/Server/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        // Check if the request has a valid token in its context.
+        if (!request.current.ctx.TryGetValue("token", out string? token) || token != _validToken)
+        {
+            Console.WriteLine($"Rejecting request with token '{token}'");
+
+            // ObjectNotExistException is one of the 6 Ice local exceptions that Ice transmits "over the wire".
+            // It usually means "no servant was found for this identity". Here, we reuse it to indicate an authorization
+            // failure without leaking information to the client.
+            throw new Ice.ObjectNotExistException();
```
bernardnormier, line 23: The most debatable aspect to this demo. There is no perfect solution with the Ice protocol since "system exceptions" are not extensible.

Defining a new user exception, like in the existing interceptor demo, is not good as it pollutes all Slice operation signatures.

UnknownException works but triggers by default a log from the logger middleware (Ice.Warn.Dispatch=1 by default). ONE or perhaps OpNE sounds like the best option.

csharp/Ice/Middleware/Client/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    string unexpected = await greeter.GreetAsync(
+        Environment.UserName,
+        context: new Dictionary<string, string> { ["token"] = "pineapple" });
+
+    Console.WriteLine($"Received unexpected greeting: {unexpected}");
+}
+catch (Ice.ObjectNotExistException)
+{
+    // Expected with an invalid token. See AuthorizationMiddleware.
```
externl, line 27:
```suggestion
    // Expected with an invalid (or missing) token. See AuthorizationMiddleware.
```

pepone (approved):

externl (approved):

PR #298, opened 2025-02-13
Add C++ build flags

externl:
This PR adds C++ build warning flags (copied from the old build system). They're on by default so that we see them when building but error on warning is not on by default. That way users don't get compilation errors if they're using slightly different compiler versions from what we test. 

In CI we enable error on warning with CMAKE's `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,11 +63,6 @@ jobs:
           working_directory: ice
           build_flags: ${{ runner.os == 'Windows' && '/t:BuildDist /p:Configuration=Release /p:Platform=x64' || 'srcs' }}
 
-      - name: Publish C# NuGet Packages
-        timeout-minutes: 5
-        working-directory: ice/csharp/msbuild
-        run: dotnet msbuild ice.proj /t:Publish /p:Configuration=Release /p:Platform=x64
-
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
--- a/cpp/cmake/common.cmake
+++ b/cpp/cmake/common.cmake
@@ -3,6 +3,25 @@ cmake_minimum_required(VERSION 3.24)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+
+if (WIN32 AND NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/packages)
+    message(STATUS "Downloading Ice NuGet package")
+    # option(ICE_NUGET_INSTALL "Use NuGet to install Ice during configuration" ON)
+    set(ICE_NUGET_SOURCE "https://download.zeroc.com/nexus/repository/nuget-nightly/" CACHE STRING "Ice NuGet package source")
+    # if (ICE_NUGET_INSTALL)
+    execute_process(
+        COMMAND nuget install -Source ${ICE_NUGET_SOURCE} -OutputDirectory ${CMAKE_CURRENT_BINARY_DIR}/packages zeroc.ice.v143 -Prerelease -ExcludeVersion
+        RESULT_VARIABLE nuget_result
+        OUTPUT_VARIABLE nuget_output
+        ERROR_VARIABLE nuget_error)
+    if (nuget_result)
+        message(FATAL_ERROR "Failed to download Ice NuGet package: ${nuget_error}")
+    endif()
+    # endif()
+
+    set(Ice_HOME "${CMAKE_CURRENT_BINARY_DIR}/packages/zeroc.ice.v143" CACHE PATH "Path to Ice installation directory")
+endif()
+
 # We use these flags over presets to avoid having to create a CMakePreset.json file in every project.
 # If you want to disable, set this option to OFF (`-DICE_DEMO_WARNINGS=OFF`) when configuring CMake.
 option(ICE_DEMO_WARNINGS "Ice Demo C++ warning flags" ON)
--- a/cpp/cmake/IceConfig.cmake
+++ b/cpp/cmake/IceConfig.cmake
@@ -23,28 +23,65 @@ if(NOT DEFINED Ice_ARCHITECTURE)
   endif()
 endif()
 
+if(EXISTS ${Ice_HOME}/cpp)
+  set(Ice_SOURCE_BUILD ON)
+endif()
+
+if(Ice_SOURCE_BUILD)
+  list(APPEND ice_include_path_suffixes "cpp/include")
+  list(APPEND ice_bin_path_suffixes "cpp/bin")
+  list(APPEND ice_lib_path_suffixes_release "cpp/lib")
+
+  if (Ice_ARCHITECTURE)
+    list(APPEND ice_lib_path_suffixes_release "cpp/lib/${Ice_ARCHITECTURE}")
+  endif()
+
+  if(WIN32)
+    list(APPEND ice_bin_path_suffixes "cpp/bin/${Ice_ARCHITECTURE}/Release" "cpp/bin/${Ice_ARCHITECTURE}/Debug")
+    list(APPEND ice_lib_path_suffixes_release "cpp/lib/${Ice_ARCHITECTURE}/Release")
+    list(APPEND ice_lib_path_suffixes_debug "cpp/lib/${Ice_ARCHITECTURE}/Debug")
+  endif()
+
+else()
+
+  list(APPEND ice_include_path_suffixes "include")
+  list(APPEND ice_bin_path_suffixes "bin")
+  list(APPEND ice_lib_path_suffixes_release "lib")
+
+  if(WIN32)
+    # NuGet packages
+    list(APPEND ice_include_path_suffixes "build/native/include")
+    list(APPEND ice_bin_path_suffixes "tools")
+    list(APPEND ice_lib_path_suffixes_release "build/native/lib/${Ice_ARCHITECTURE}/Release")
+    list(APPEND ice_lib_path_suffixes_debug "build/native/lib/${Ice_ARCHITECTURE}/Debug")
+  endif()
+
+endif()
+
 # Ice include directory
 if(NOT DEFINED Ice_INCLUDE_DIR)
-  find_path(Ice_INCLUDE_DIR NAMES Ice/Ice.h HINTS ${Ice_HOME}/cpp/include CACHE PATH "Path to the Ice include directory")
+  find_path(Ice_INCLUDE_DIR NAMES Ice/Ice.h HINTS ${Ice_HOME} PATH_SUFFIXES ${ice_include_path_suffixes} CACHE PATH "Path to the Ice include directory")
 endif()
 
 # Ice include directories include the generated directory(s)
 if(NOT DEFINED Ice_INCLUDE_DIRS)
   set(Ice_INCLUDE_DIRS ${Ice_INCLUDE_DIR})
 
-  #  Only the Windows build has separate Debug and Release directories for the generated files
-  if(WIN32)
-    set(Ice_GENERATED_INCLUDE_DIR_RELEASE ${Ice_INCLUDE_DIR}/generated/${Ice_ARCHITECTURE}/Release)
-    if (EXISTS ${Ice_GENERATED_INCLUDE_DIR_RELEASE})
-      list(APPEND Ice_INCLUDE_DIRS $<$<CONFIG:Release>:${Ice_GENERATED_INCLUDE_DIR_RELEASE}>)
-    endif()
+  if(Ice_SOURCE_BUILD)
+    #  Only the Windows build has separate Debug and Release directories for the generated files
+    if(WIN32)
+      set(Ice_GENERATED_INCLUDE_DIR_RELEASE ${Ice_INCLUDE_DIR}/generated/${Ice_ARCHITECTURE}/Release)
+      if (EXISTS ${Ice_GENERATED_INCLUDE_DIR_RELEASE})
+        list(APPEND Ice_INCLUDE_DIRS $<$<CONFIG:Release>:${Ice_GENERATED_INCLUDE_DIR_RELEASE}>)
+      endif()
 
-    set(Ice_GENERATED_INCLUDE_DIR_DEBUG ${Ice_INCLUDE_DIR}/generated/${Ice_ARCHITECTURE}/Debug)
-    if (EXISTS ${Ice_GENERATED_INCLUDE_DIR_DEBUG})
-      list(APPEND Ice_INCLUDE_DIRS $<$<CONFIG:Debug>:${Ice_GENERATED_INCLUDE_DIR_DEBUG}>)
+      set(Ice_GENERATED_INCLUDE_DIR_DEBUG ${Ice_INCLUDE_DIR}/generated/${Ice_ARCHITECTURE}/Debug)
+      if (EXISTS ${Ice_GENERATED_INCLUDE_DIR_DEBUG})
+        list(APPEND Ice_INCLUDE_DIRS $<$<CONFIG:Debug>:${Ice_GENERATED_INCLUDE_DIR_DEBUG}>)
+      endif()
+    else()
+        list(APPEND Ice_INCLUDE_DIRS ${Ice_INCLUDE_DIR}/generated)
     endif()
-  elseif(EXISTS ${Ice_INCLUDE_DIR}/generated)
-      list(APPEND Ice_INCLUDE_DIRS ${Ice_INCLUDE_DIR}/generated)
   endif()
 
   set(Ice_INCLUDE_DIRS ${Ice_INCLUDE_DIRS} CACHE STRING "Ice include directories")
@@ -53,7 +90,6 @@ endif()
 
 # Read Ice version variables from Ice/Config.h
 if (NOT DEFINED Ice_VERSION)
-
   file(STRINGS "${Ice_INCLUDE_DIR}/Ice/Config.h" _ice_config_h_content REGEX "#define ICE_([A-Z]+)_VERSION ")
 
   if("${_ice_config_h_content}" MATCHES "#define ICE_STRING_VERSION \"([^\"]+)\"")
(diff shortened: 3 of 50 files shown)


cpp/cmake/common.cmake
```diff
@@ -1,8 +1,21 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.24)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# We use these flags over presets to avoid having to create a CMakePreset.json file in every project.
+# If you want to enable these, set this option to OFF: -DICE_DEMO_WARNINGS=OFF
```
bernardnormier, line 7: You mean _disable_ these warnings?

bernardnormier:
Something is wrong with the Windows CI build. It succeeds but the log shows errors:

```
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.42.34433\bin\HostX64\x64\CL.exe /c /I"D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\generated\writer" /nologo /W1 /WX /diagnostics:column /O2 /Ob2 /D _MBCS /D WIN32 /D _WINDOWS /D NDEBUG /D "CMAKE_INTDIR=\"Release\"" /EHsc /MD /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /std:c++17 /Fo"writer.dir\Release\\" /Fd"writer.dir\Release\vc143.pdb" /external:W0 /Gd /TP /errorReport:queue  /external:I "D:/a/ice-demos/ice-demos/ice/cpp/include" /external:I "D:/a/ice-demos/ice-demos/ice/cpp/include/generated/x64/Release" "D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp" "D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\generated\writer\Stock.cpp"
  Writer.cpp
  Stock.cpp
Error: D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(72,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
Error: D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(81,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
Error: D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(90,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
Error: D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(99,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
Error: D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(108,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
  Generating Code...
Done Building Project "D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj" (default targets) -- FAILED.
Done Building Project "D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\ALL_BUILD.vcxproj" (default targets) -- FAILED.

Build FAILED.

"D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\ALL_BUILD.vcxproj" (default target) (1) ->
"D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj" (default target) (4) ->
(ClCompile target) -> 
  D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(72,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
  D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(81,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
  D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(90,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
  D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(99,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]
  D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\Writer.cpp(108,18): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ice-demos\ice-demos\ice-demos\cpp\DataStorm\stock\build\writer.vcxproj]

    0 Warning(s)
    5 Error(s)

Time Elapsed 00:00:09.76
```

Once this issue is fixed (build error => build fails), you should fix these errors by building this particular demo with c++20.

cpp/cmake/common.cmake
externl: Yes, I went back and forth a few times on what the default should be.

cpp/DataStorm/stock/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ project(datastorm_stock CXX)
 
 include(../../cmake/common.cmake)
 
+# This demo requires C++20
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
pepone, line 9: I would keep it C++17 compatible, using std::ranges is not that important for the demo.

cpp/Ice/config/Client.cpp
```diff
@@ -12,11 +12,14 @@ int
 main(int argc, char* argv[])
 {
     // Figure out my name.
+#ifdef _MSC_VER
+    size_t requestedSize;
+    char buffer[256];
+    getenv_s(&requestedSize, buffer, sizeof(buffer), "USERNAME");
```
pepone, line 18: in theory we should check the return value for this function. It can return an error...

cpp/Ice/context/Client.cpp
```diff
@@ -11,19 +11,23 @@ int
 main(int argc, char* argv[])
 {
     // Figure out my name.
+#ifdef _MSC_VER
```
pepone, line 14: I find this updated getenv gets in the way of the sample. I would rather keep the previous code and disable the warning, or use a fixed name.

cpp/IceStorm/clock/Publisher.cpp
```diff
@@ -148,10 +148,26 @@ main(int argc, char* argv[])
             while (true)
             {
                 auto now = chrono::system_clock::to_time_t(chrono::system_clock::now());
+                std::tm timeInfo;
+
+#ifdef _MSC_VER
```
pepone, line 153: I find this also gets in the way of the demo, can we just keep the previous code with a disabled warning.

pepone (approved):

cpp/DataStorm/stock/CMakeLists.txt
externl: We need c++20 for other parts of this demo, namely it was already using designated initializer syntax. I think its fine for a few demos to use c++20

cpp/Ice/context/Client.cpp
externl: ok, I'll will undo and disable the warning

cpp/IceStorm/clock/Publisher.cpp
externl: The issue with the previous code is that `localtime` is not thread safe. Are we sure all of these callbacks are being called serially?

externl:
I was talking with @bernardnormier and he suggested to put the time conversion and username things a `common` folder.

cpp/DataStorm/stock/CMakeLists.txt
pepone: I guess is fine if it can be build in all supported platforms.

.github/workflows/ci.yml
```diff
@@ -118,9 +128,12 @@ jobs:
         timeout-minutes: 20
         working-directory: ice-demos/csharp
         run: |
+          $ErrorActionPreference = "Stop"
           Get-ChildItem -Recurse -Filter *.sln | ForEach-Object {
             dotnet build $_.FullName
+           if ($LASTEXITCODE -ne 0) { throw "CMake configuration failed" }
```
pepone, line 134:
```suggestion
           if ($LASTEXITCODE -ne 0) { throw "dotnet build failed" }
```

cpp/IceGrid/simple/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ include(../../cmake/common.cmake)
 
 add_executable(client Client.cpp Hello.ice)
 slice2cpp_generate(client)
-target_link_libraries(client Ice::Ice Ice::IceGrid)
+target_link_libraries(client Ice::Ice Ice::IceGrid Ice::Glacier2)
```
pepone, line 9: How does this work? do we still rely on `#pragma comment`?

cpp/IceStorm/clock/Subscriber.cpp
```diff
@@ -168,7 +168,7 @@ main(int argc, char* argv[])
     // Add a servant to the object adapter. If --id is used the identity comes from the command line, otherwise a UUID
     // is used. id is not directly altered since it is used below to detect whether subscribeAndGetPublisher can throw
     // AlreadySubscribed.
-    Ice::Identity subId{.name = id.empty() ? Ice::generateUUID() : id};
+    Ice::Identity subId{id.empty() ? Ice::generateUUID() : id, ""};
```
pepone, line 171: Is category `""` required? I thought Bernard recently fixed this.

cpp/common/time.h
```diff
@@ -0,0 +1,38 @@
...
+    /// @return The formatted time point.
+    std::string formatTime(std::chrono::system_clock::time_point timePoint)
+    {
+        std::time_t time = std::chrono::system_clock::to_time_t(timePoint);
+        std::tm timeInfo;
+
+#ifdef _MSC_VER
+        if (localtime_s(&timeInfo, &time))
+        {
+            std::cerr << "localtime_s error" << std::endl;
```
pepone, line 19: We can throw `runtime_error` it would be very rare that this fails but still seems clear.

pepone (approved):

cpp/IceGrid/simple/CMakeLists.txt
externl: No, we don't rely on pragma comment. It doesn't cause any issues so we don't worry about it.

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) ZeroC, Inc.
+
+#include <optional>
+#include <string>
+
+namespace common
+{
+    /// Get the value of an environment variable.
```
bernardnormier, line 8: Gets

cpp/IceStorm/clock/Subscriber.cpp
externl: We can't use designed initializer syntax  as it's c++20 and windows would complain. Bernard did recently fix it but it  was really for GCC. Windows won't let you do it.

cpp/common/time.h
externl: I'd prefer to just print then return an empty string. Otherwise we'd need to worry a little about exception handling.

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) ZeroC, Inc.
+
+#include <optional>
+#include <string>
+
+namespace common
```
bernardnormier, line 6: I would call it `Env`, not common. With a PascalName namespace since it's our convention.

Same for the file => Env.h.

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// Get the value of an environment variable.
+    /// @param variableName The name of the environment variable.
+    /// @return The value of the environment variable, or nullopt if the variable is not set.
+    std::optional<std::string> getEnvironmentVariable(const char* variableName)
+    {
+#ifdef _MSC_VER
```
bernardnormier, line 13: I think _WIN32 would be more appropriate: as I understand it, `getenv_s` is a Windows API, not a VC++ API.

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) ZeroC, Inc.
+
+#include <optional>
```
bernardnormier, line 3: You should include `<cstdlib>` for getenv

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
...
+        const char* value = getenv(variableName);
+#endif
+        return value ? std::optional<std::string>(value) : std::nullopt;
+    }
+
+    /// Get the username of the current user.
+    /// @return The username of the current user.
+    std::string getUsername()
+    {
+#ifdef _MSC_VER
```
bernardnormier, line 32: This should be _WIN32

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
...
+        const auto name = common::getEnvironmentVariable("USERNAME");
```
bernardnormier, line 33: Without the optional, this const would disable NVRO. I'd drop the optional (see above).

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <optional>
+#include <string>
+
+namespace common
+{
+    /// Get the value of an environment variable.
+    /// @param variableName The name of the environment variable.
+    /// @return The value of the environment variable, or nullopt if the variable is not set.
+    std::optional<std::string> getEnvironmentVariable(const char* variableName)
```
bernardnormier, line 11: I think it would be simpler to return a string, and not distinguish between "not set" and "set to empty".

cpp/common/env.h
```diff
@@ -0,0 +1,39 @@
...
+    {
+#ifdef _MSC_VER
+        size_t requestedSize;
+        char buffer[256];
+        if (getenv_s(&requestedSize, buffer, sizeof(buffer), variableName))
+        {
+            // getenv_s failed, set requestedSize to 0 to indicate that the buffer is not valid.
```
bernardnormier, line 18: I am not fond of this code, looks hacky.

cpp/common/time.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) ZeroC, Inc.
+
+#include <chrono>
+#include <iostream>
+#include <string>
+
+namespace common
```
bernardnormier, line 7: namespace Time in file Time.h

cpp/common/time.h
bernardnormier: I agree, I am not fond of reporting errors via cerr.

cpp/common/time.h
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// Format a time point as a string.
+    /// @return The formatted time point.
+    std::string formatTime(std::chrono::system_clock::time_point timePoint)
+    {
+        std::time_t time = std::chrono::system_clock::to_time_t(timePoint);
+        std::tm timeInfo;
+
+#ifdef _MSC_VER
```
bernardnormier, line 16: Should be _WIN32.

cpp/IceStorm/replicated2/Publisher.cpp
```diff
@@ -1,10 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
-#ifdef _MSC_VER
-// For localtime warning
-#    define _CRT_SECURE_NO_WARNINGS
-#endif
-
+#include "../../common/time.h"
```
bernardnormier, line 3: I would prefer a simpler `#include "Time.h"`

cpp/IceStorm/clock/Subscriber.cpp
bernardnormier: The issue here is we're building in C++17 mode on all platforms and VC++ doesn't support designated initializers in C++17 mode.

cpp/Ice/greeter/Client.cpp
```diff
@@ -11,16 +12,8 @@ using namespace std;
...
-    const char* name = getenv("USER");
-    if (name == nullptr)
-    {
-        name = getenv("USERNAME");
-    }
-    if (name == nullptr)
-    {
-        name = "masked user";
-    }
+    // Helper function to get the username of the current user.
```
bernardnormier, line 15: I would get rid of this name variable, we don't use it in other languages.

In the first greet, just call:
```cpp
string greeting = greeter->greet(Env::getUsername());
```

cpp/common/env.h
externl: All the env APIs make this distinction. I don't find empty string to be any "simpler".

cpp/common/env.h
externl: I'd prefer to keep the optional.

cpp/IceStorm/replicated2/Publisher.cpp
externl: I rater use "../../common". It's where the file is and I'd rather not add an include directory to each target, makes the cmake file uglier.

cpp/common/env.h
externl: Fixed.

cpp/common/env.h
externl: Fixed.

cpp/common/env.h
externl: Fixed.

cpp/common/env.h
externl: Fixed.

cpp/common/env.h
externl: I like the optional. And we use it in a demo.

cpp/common/env.h
externl: See comment above.

cpp/common/env.h
externl: Fixed.

cpp/common/time.h
externl: Fixed.

PR #305, opened 2025-02-19
Upgrade C# demos to use Nightly builds

pepone:
This PR upgrades C# builds to use the Nightly builds by default.

The `IceVersion` MSBuild property defined in `csharp/Directory.Build.props`  allows to override what version to use. See details in `csharp/README.md`

The `csharp/nuget.config` includes the nightly source to make the packages available. I added a package pattern so that `zeroc` prefixes packages are lookup only in this feed. This affects `zeroc.icebuilder.msbuild` used by the samples too.

One issue is that nightly uses a clean policy to delete packages older than 5 days, and this would require to re upload the builder ...

### DIFF
--- a/csharp/README.md
+++ b/csharp/README.md
@@ -26,7 +26,7 @@ Each subdirectory includes a small demo focusing on a specific Ice component or 
 
 The top-level [Directory.Build.props](./Directory.Build.props) file sets `IceVersion` to
 `3.8.0-nightly.*` by default to pull the latest nightly build of Ice.
-You can override this on the command line or in your environment:
+You can override this on the command line:
 
 ```shell
 dotnet build -p:IceVersion="3.8.0~alpha0"
--- a/csharp/README.md
+++ b/csharp/README.md
@@ -2,64 +2,34 @@
 
 - [C# Demos](#c-demos)
   - [Overview](#overview)
-  - [Building and running the Demos](#building-and-running-the-demos)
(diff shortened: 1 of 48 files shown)


pepone:
Seems the Nightly package only includes the Windows compiler. Need to fix that before merging this pr.

externl (approved):

csharp/README.md
```diff
@@ -2,64 +2,34 @@
...
-Open a command prompt and change to the `csharp` subdirectory:
-
-```shell
-cd csharp
-```
-
-To build the sample programs run:
+The top-level [Directory.Build.props](./Directory.Build.props) file sets `IceVersion` to
+`3.8.0-nightly.*` by default to pull the latest nightly build of Ice.
+You can override this on the command line or in your environment:
```
bernardnormier, line 29: We only show command-line so I would not mention "in your environment". And given the code above, I don't see how you can change this version with an env variable.

Then, how do you see it working once we release? `main` uses the latest nightly while the 3.8 branch uses the latest released 3.8.x?

bernardnormier (approved):

csharp/README.md
pepone, line 29:
```suggestion
You can override this on the command line:
```

pepone:
> ci.yml should be updated to no build Ice for C#

Remove the publishing of local NuGet. We still build everything with the top-level MSbuild file.

PR #306, opened 2025-02-19
Port new bidir demo to JavaScript

pepone:
This PR brings the new bidir demo to JavaScript.

It depends on the BigInt updates from https://github.com/zeroc-ice/ice/pull/3574


js/Ice/bidir/client.ts
```diff
@@ -0,0 +1,40 @@
...
+console.log("Stop button pressed, exiting...");
+
+const millisecondsBeforeEpoch = 719162n * 24n * 60n * 60n * 1000n; // daysBeforeEpoch converted to milliseconds
+const ticksPerMillisecond = 10000n; // number of ticks per millisecond
+
+// Convert a timestamp representing the number of milliseconds elapsed since the epoch to a BigInt representing
+// the number of ticks (100 nanoseconds) since January 1, 0001 00:00:00 UTC in the Gregorian calendar.
+function toTicks(timestamp: number) {
+    return (BigInt(timestamp) + millisecondsBeforeEpoch) * ticksPerMillisecond;
+}
```
externl, line 40: Should we move this into a separate file.

js/Ice/bidir/client.ts
```diff
@@ -0,0 +1,40 @@
...
+console.log("Wake-up call scheduled, falling asleep...");
+
+// Wait until the "stop" button is pressed on the mock alarm clock.
+await mockAlarmClock.StopPressed;
+console.log("Stop button pressed, exiting...");
+
+const millisecondsBeforeEpoch = 719162n * 24n * 60n * 60n * 1000n; // daysBeforeEpoch converted to milliseconds
+const ticksPerMillisecond = 10000n; // number of ticks per millisecond
+
+// Convert a timestamp representing the number of milliseconds elapsed since the epoch to a BigInt representing
```
externl, line 36: Why don't we use the standard UNIX epoch of `January 1, 1970,`. It's the default in many languages or they generally have a UnixEpoch function.

externl (approved):

js/Ice/bidir/client.ts
pepone: Here the JavaScript client has to use the format expected by the server.

For the servers we pick this representation because it matches IceRPC timestamp type. https://github.com/icerpc/icerpc-csharp/blob/10f9e641e0915d2a832edf19aeb3fcca801780f3/slice/WellKnownTypes/TimeStamp.slice#L8

It is also the same representation used by protobuf https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp

js/Ice/bidir/MockAlarmClock.ts
```diff
@@ -0,0 +1,30 @@
+
+import { Ice } from "@zeroc/ice";
+import { EarlyRiser } from "./AlarmClock.js";
+
+export class MockAlarmClock extends EarlyRiser.AlarmClock
+{
+    private _needMoreTime = true;
+    private _resolver: ((value: void | PromiseLike<void>) => void);
```
bernardnormier, line 8: Is this a standard name/pattern in JS/TS?

We have a Utils.Resolver in src, but that's not what you're using here.

js/Ice/bidir/MockAlarmClock.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    public StopPressed = new Promise<void>((resolve, reject) => this._resolver = resolve);
+
```
bernardnormier, line 11: Extra blank line

js/Ice/bidir/MockAlarmClock.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    // Implements the abstract method ring from the AlarmClock class generated by the Slice compiler.
```
bernardnormier, line 13: I would add a comment like "can return a value or a future" since we don't have other demos showing off async servant methods.

js/Ice/bidir/client.ts
```diff
@@ -0,0 +1,40 @@
...
+// Register the MockAlarmClock servant with the adapter. The wake up service knows we use identity "alarmClock".
+var mockAlarmClock = new MockAlarmClock();
+adapter.add(mockAlarmClock, Ice.stringToIdentity("alarmClock"));
+
+// Create a proxy to the wake-up service.
+const wakeUpService = new EarlyRiser.WakeUpServicePrx(communicator, "wakeUpService:tcp -h localhost -p 4061");
+
+// Schedule a wake-up call in 5 seconds. This call establishes the connection to the server; incoming requests over this
+// connection are handled by the communicator's default object adapter.
+
```
bernardnormier, line 25: extra blank line

bernardnormier (approved):

js/Ice/bidir/MockAlarmClock.ts
pepone: the standard promise takes to callbacks resolve and reject.

js/Ice/bidir/MockAlarmClock.ts
pepone: I updated the return type to be `EarlyRiser.ButtonPressed | Promise<EarlyRiser.ButtonPressed>`

Which matches what we generate in the TypeScript definition.

bernardnormier: I'd rather only return EarlyRiser.ButtonPressed and update the comment to be:
```
// Implements the abstract method ring from the AlarmClock class generated by the Slice compiler.
// Here, we provide a synchronous implementation that returns a value (EarlyRiser.ButtonPressed). We
// could also provide an async implementation that returns a promise (Promise<EarlyRiser.ButtonPressed> for
// this method).
```

externl: using `EarlyRiser.ButtonPressed | Promise<EarlyRiser.ButtonPressed>` is very typescript

bernardnormier: ok

PR #309, opened 2025-02-20
Add Python greeter demo

pepone:
This PR ports the Greeter demo to Python and establishes the structure for future Python demos.


python/Ice/greeter/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Running the server
+
+Create and activate a Python virtual environment in the `server` or `server_amd` directory to run the synchronous or asynchronous version of the server application, respectively.
+
+From the activated virtual environment in the `server` or `server_amd` directory, install the program dependencies and run the application using the following command:
+
+```shell
+pip install -r requirements.txt
+python3 main.py
```
externl, line 17: I think should be consistent with usage of `pip/python` and `pip3/python3`. The python docs seem to prefer to just use `pip` and `python` these days.

python/Ice/greeter/server_amd/chatbot.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Chatbot(VisitorCenter.Greeter):
+
+    def __init__(self, loop):
+        self.loop = loop
+
+    def greet(self, name, current):
+        return asyncio.run_coroutine_threadsafe(self.slow_greet(name), self.loop)
+
+    async def slow_greet(self, name):
```
externl, line 20: we should use python3 type hinting

externl (approved):

python/Ice/greeter/README.md
pepone: not sure if all platforms provide python alias...

externl: They do inside the virtualenv. If you look at the doc page you link they use `python` exclusively

bernardnormier (commented):
My understanding is you can use AMD (and async dispatches) _without_ asyncio, and unfortunately the server_amd gives the impression AMD and asyncio are tied.

python/Ice/greeter/server/chatbot.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) ZeroC, Inc.
+
+import Ice
+
+# Load the Slice file that defines the Greeter interface.
+Ice.loadSlice('../slice/Greeter.ice')
```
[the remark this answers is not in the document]
pepone: I will switch to generated code with slice2py.

pepone:
> My understanding is you can use AMD (and async dispatches) without asyncio, and unfortunately the server_amd gives the impression AMD and asyncio are tied.

According to the doc AMD is not tied to asyncio

https://doc.zeroc.com/ice/3.7/language-mappings/python-mapping/server-side-slice-to-python-mapping/asynchronous-method-dispatch-amd-in-python

But my understanding is that we still need an event loop

python/Ice/greeter/server/main.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+# Copyright (c) ZeroC, Inc.
+
+import Ice
+import chatbot
+import signal
+import sys
+import threading
+
+exit_event = threading.Event()
```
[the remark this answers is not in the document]
pepone: Ok. I'm not sure if this is the correct approach in Python. `set` might acquire a lock under the hood.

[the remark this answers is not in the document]
bernardnormier: Ok, we should review this code in a follow-up PR.

python/Ice/greeter/server_amd/chatbot.py
```diff
@@ -0,0 +1,20 @@
...
+# Slice module VisitorCenter in Greeter.ice maps to Python module VisitorCenter.
+import VisitorCenter
+
+class Chatbot(VisitorCenter.Greeter):
+
+    def __init__(self, loop):
+        self.loop = loop
+
+    def greet(self, name: str, current: Ice.Current) -> str:
+        return asyncio.run_coroutine_threadsafe(self.slow_greet(name), self.loop)
```
bernardnormier, line 15: We need comments here.

bernardnormier (approved):
We need more comments in this demo before using it as a model for other demos.

PR #311, opened 2025-02-21
Use Linux and macOS nightly repos

externl:
This PR updates the Linux and macOS builds to use our nightly brew and apt packaging. 

For now I've disable the Java build as we want to fix this demos anyway and they still rely on the nightly.


pepone (approved):
Looks good.

.github/actions/install-nightly-ice/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Install Ice APT Repository
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo mkdir -p /etc/apt/keyrings
+        sudo curl -fsSL https://download.zeroc.com/GPG-KEY-zeroc-nightly | sudo gpg --dearmor -o /etc/apt/keyrings/zeroc-nightly.gpg
+        echo "deb [signed-by=/etc/apt/keyrings/zeroc-nightly.gpg] https://download.zeroc.com/nexus/repository/ubuntu24.04-nightly nightly main" | sudo tee /etc/apt/sources.list.d/zeroc-nightly.list
+
+        sudo apt-get update
+        sudo apt-get install zeroc-ice-all-dev
```
pepone, line 22: don't we need to pass `-y` for non interactive mode?

pepone (approved):

swift/Ice/greeter/Package.swift
```diff
@@ -3,28 +3,24 @@
...
-    dependencies: [.package(name: "ice", path: iceHome)],
+    dependencies: [.package(url: "https://download.zeroc.com/nexus/repository/nightly/ice-swift.git", branch: "master")],
     targets: [
         .executableTarget(
             name: "Client",
-            dependencies: [.product(name: "Ice", package: "ice")],
+            dependencies: [.product(name: "Ice", package: "ice-swift")],
             exclude: ["slice-plugin.json"],
-            plugins: [.plugin(name: "CompileSlice", package: "ice")]
+            plugins: [.plugin(name: "CompileSlice", package: "ice-swift")]
```
pepone, line 17: Can't we use 'ice' for the package name?

pepone (approved):

.github/actions/install-nightly-ice/action.yml
externl: I think they set some env variables so you don't need do but I think it's better to add it.

PR #312, opened 2025-02-21
Add Java greeter demo

pepone:
This PR is the initial draft for Greeter demo port of Java.


java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
```diff
@@ -0,0 +1,37 @@
...
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.Util;
+import com.zeroc.greeter.VisitorCenter.GreeterPrx;
+import java.util.concurrent.CompletableFuture;
+
+class Client {
+    public static void main(String[] args)
+    {
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+        try(com.zeroc.Ice.Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```
bernardnormier, line 14: Is this formatting correct? I prefer `try (` and `catch (` to distinguish with method calls.

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
```diff
@@ -0,0 +1,37 @@
...
+            // or IP address.
+            GreeterPrx greeter = GreeterPrx.createProxy(communicator, "greeter:tcp -h localhost -p 4061");
+
+            // Send a request to the remote object and get the response.
+            String greeting = greeter.greet(System.getProperty("user.name"));
+            System.out.println(greeting);
+
+            // Send another request to the remote object, this time with greetAsync. greetAsync returns a future
+            // immediately.
+            CompletableFuture<String> result = greeter.greetAsync("alice");
```
bernardnormier, line 28: It would call it futureGreeting like in the C++ demo.

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
```diff
@@ -0,0 +1,37 @@
...
+
+            try {
+                System.out.println(result.get());
+            } catch(Exception ex) {
```
bernardnormier, line 32: Unless it's required, no try/catch block just like in the C++ demo:
```java
greeting = futureGreeting.get();
System.out.println(greeting);
```

java/Ice/greeter/client/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    java
+    application
+    id("com.zeroc.gradle.ice-builder.slice") version "1.5.1"
+}
+
+dependencies {
+    implementation("com.zeroc:ice:3.8.0-nightly-+")
+}
+
+slice {
```
bernardnormier, line 11: I assume we can include comments in these files?

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
pepone: We need to handle `java.lang.InterruptedException` and `java.util.concurrent.ExecutionException`

java/Ice/greeter/client/build.gradle.kts
pepone: Yes I will add some comments here.

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
pepone: Maybe I should configure spotless plug-in for the demos too?

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.greeter;
+
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.Util;
+import com.zeroc.greeter.VisitorCenter.GreeterPrx;
+import java.util.concurrent.CompletableFuture;
+
+class Client {
```
InsertCreativityHere, line 10: I feel like this should probably be public, since it's not an internal implementation detail or anything.
In Java, putting no access qualified on a class is equivalent to `internal` in C#.
```suggestion
public class Client {
```

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
InsertCreativityHere, line 14: Indeed, there should be a space here.

Looking at other demos though, it seems like the demos use our old Java style,
before we switched to all these other spotless styles.

There's also no need to qualify `Communicator` since it's imported.
I personally would leave the qualification on `Util`, since `util` is very common, even in the Java Standard Library.
```suggestion
        try (Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```

java/Ice/greeter/client/src/main/java/com/zeroc/greeter/Client.java
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.greeter;
+
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.Util;
```
InsertCreativityHere, line 6: If you decide to leave the `Util` call qualified, should also remove this from the `import` list.
In Java, `import`s are just glorified `using`.
```suggestion
```

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Chatbot.java
```diff
@@ -0,0 +1,20 @@
...
+
+package com.zeroc.greeter;
+
+import com.zeroc.greeter.VisitorCenter.Greeter;
+import com.zeroc.Ice.Current;
+
+/**
+ * Chatbot is an Ice servant that implements Slice interface Greeter.
+ */
+class Chatbot implements Greeter
```
InsertCreativityHere, line 11: Also feels public to me.
```suggestion
public class Chatbot implements Greeter
```

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Server.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.greeter;
+
+import com.zeroc.Ice.Communicator;
```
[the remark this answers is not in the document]
InsertCreativityHere: I would lean towards keeping these imports and removing the unnecessary qualification below.
In Java, it's pretty common to have lots of imports, and keep your actual code more readable.

Except I'd leave the qualification on `Util`, since it's an ambiguous name, and there are `util`s in the standard library. So it's good to keep it clear it for readers I think.

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Server.java
```diff
@@ -0,0 +1,35 @@
...
+import com.zeroc.Ice.Util;
+import com.zeroc.greeter.VisitorCenter.GreeterPrx;
+import com.zeroc.greeter.Chatbot;
+
+class Server {
+    public static void main(String[] args)
+    {
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+        try(com.zeroc.Ice.Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```
InsertCreativityHere, line 14:
```suggestion
        try (Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```

java/Ice/greeter/serveramd/src/main/java/com/zeroc/greeter/Chatbot.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.zeroc.greeter.VisitorCenter.Greeter;
+import com.zeroc.Ice.Current;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Chatbot is an Ice servant that implements Slice interface Greeter.
+ */
+class Chatbot implements Greeter
```
InsertCreativityHere, line 13: Same
```suggestion
public class Chatbot implements Greeter
```

java/Ice/greeter/serveramd/src/main/java/com/zeroc/greeter/Server.java
```diff
@@ -0,0 +1,35 @@
...
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.Util;
+import com.zeroc.greeter.VisitorCenter.GreeterPrx;
+import com.zeroc.greeter.Chatbot;
+
+class Server {
+    public static void main(String[] args)
+    {
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+        try(com.zeroc.Ice.Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```
InsertCreativityHere, line 14: Same
```suggestion
        try (Communicator communicator = com.zeroc.Ice.Util.initialize(args))
```

InsertCreativityHere (approved):
Looks good, just some superficial things.

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Chatbot.java
pepone: We use internal for C#

```
internal class Chatbot : GreeterDisp_
```

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Chatbot.java
```diff
@@ -0,0 +1,20 @@
...
+{
+    // Implements the abstract method Greet from the Greeter interface generated by the Slice compiler.
```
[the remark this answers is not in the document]
pepone: What do you mean by it works, that the generated doc for Chatbot includes the inherited doc?

java/Ice/greeter/serveramd/src/main/java/com/zeroc/greeter/Server.java
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                communicator.shutdown();
```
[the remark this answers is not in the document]
pepone: Yes, we need to decide what pattern to keep.

java/Ice/greeter/README.md
```diff
@@ -0,0 +1,43 @@
...
+- `server`: A synchronous dispatch implementation
+- `serveramd`: An asynchronous dispatch implementation using Ice's Asynchronous Method Dispatch (AMD)
+
+The same client works with both.
+
+## Building the Project
+
+This project uses Gradle and the [Application plugin]. To build the client and server applications, run:
+
+```shell
```
bernardnormier, line 15: It would be nice to provide the Windows version too.

java/Ice/greeter/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This demo provides two implementations for the server:
+- `server`: A synchronous dispatch implementation
+- `serveramd`: An asynchronous dispatch implementation using Ice's Asynchronous Method Dispatch (AMD)
+
+The same client works with both.
+
+## Building the Project
+
+This project uses Gradle and the [Application plugin]. To build the client and server applications, run:
```
bernardnormier, line 13: I assume this is a reasonably common thing to do?

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -0,0 +1,35 @@
...
+package com.zeroc.demos.greeter;
+
+import com.zeroc.demos.greeter.VisitorCenter.GreeterPrx;
+import com.zeroc.Ice.Communicator;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+class Client {
+    public static void main(String[] args) {
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
```
bernardnormier, line 12: The lines are > 120 columns.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -0,0 +1,35 @@
...
+            // Send a request to the remote object and get the response.
+            String greeting = greeter.greet(System.getProperty("user.name"));
+            System.out.println(greeting);
+
+            // Send another request to the remote object, this time with greetAsync. greetAsync returns a future
+            // immediately.
+            CompletableFuture<String> futureGreeting = greeter.greetAsync("alice");
+
+            try {
+                System.out.println(futureGreeting.get());
```
bernardnormier, line 29: I would prefer `greeting = futureGreeting.get();`

Not sure about the exception handling. Could we catch only checked exceptions here?

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
```diff
@@ -0,0 +1,32 @@
...
+import com.zeroc.demos.greeter.Chatbot;
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.ObjectAdapter;
+
+class Server {
+    public static void main(String[] args) {
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+        try (Communicator communicator = com.zeroc.Ice.Util.initialize(args)) {
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                communicator.shutdown();
```
bernardnormier, line 14: This is not correct since the shutdown hook terminates the application when it completes.

bernardnormier (approved):

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Chatbot.java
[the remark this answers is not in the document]
bernardnormier: Yes. And that the doc-comment on the Slice operation is correctly generated on the generated Greeter Java method.

java/Ice/greeter/slice/Greeter.ice
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+[["java:package:com.zeroc.greeter"]]
```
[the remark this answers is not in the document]
bernardnormier: I was suggesting `com.zeroc.demos` for the Slice file (=> maps the com.zeroc.demos.VisitorCenter) and `com.zeroc.demos.greeter` to the hand-written demo code.

Putting the generated code in com.zeroc.demos.greeter.VisitorCenter provides no benefit: a sub-package is just a different package in Java. The visibility is the same as a sibling package.

Then, once we can remap the package names, I would _not_ remap VisitorCenter => greeter. It would reduce some qualification but make the setup more confusing. And naturally, the generated code is mostly public: its user doesn't need to be and usually isn't in the same package.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
bernardnormier: What about:
```java
            try {
                greeting = futureGreeting.get();
            } catch (InterruptedException e) {
                assert false; // We don't interrupt this thread.
                return;
            }
            catch (ExecutionException e) {
                System.out.println("Could not get greeting: " + e.getMessage());
                return;
            }

            System.out.println(greeting);
```

This is longer but clearer in my view.

java/Ice/greeter/server/src/main/java/com/zeroc/greeter/Chatbot.java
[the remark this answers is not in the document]
pepone: The generated doc comment looks correct.
```
    /**
     * Creates a personalized greeting.
     * @param name The name of the person to greet.
     * @param current The Current object for the invocation.
     * @return The greeting.
     **/
    String greet(String name, com.zeroc.Ice.Current current);
```

java/Ice/greeter/slice/Greeter.ice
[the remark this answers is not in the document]
pepone: maybe we can remap VistorCenter to vistorcenter once this is supported? the package names are always lower case by convention.

java/Ice/greeter/README.md
pepone: That is what `gradle init` does when you create a new Java Application project.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
pepone: Agree, it is much clear.

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
pepone: Updated to use destroy. We should fix this in a follow up PR to used the new proposed pattern.

java/Ice/greeter/README.md
pepone: Added Windows instructions

PR #321, opened 2025-02-26
Update greeter async server to use the new coroutine executor

pepone:
This PR updates the async greeter server to use the new coroutine executor.

server_amd was renamed to server_asyncio to make it clear that is using asyncio.


python/Ice/greeter/server_asyncio/main.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def coroutineExecutor(coroutine):
+        """
+        A custom coroutine executor that runs the coroutine in the main asyncio event loop.
+        """
+
+        # We use run_coroutine_threadsafe because the executor is called by an Ice thread.
+        return asyncio.run_coroutine_threadsafe(coroutine, loop)
+
+    # Create an initialization data structure that specifies the coroutine executor.
```
bernardnormier, line 21: I would replace "specifies" by simply "sets" or maybe "configures".

python/Ice/greeter/server_asyncio/main.py
```diff
@@ -0,0 +1,46 @@
...
+    initData = Ice.InitializationData()
+    initData.coroutineExecutor = coroutineExecutor
+
+    # Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
```
bernardnormier, line 25: "disposed" is a C# term, and the scope is not the program but the `with block`. What about:

> The communicator is destroyed at the end of the with block.

python/Ice/greeter/server_asyncio/main.py
```diff
@@ -0,0 +1,46 @@
...
+    with Ice.initialize(initData) as communicator:
+
+        #print("Communicator initialized", communicator.getCoroutineExecutor())
```
bernardnormier, line 28: Debugging letfover?

python/Ice/greeter/server_asyncio/main.py
```diff
@@ -0,0 +1,46 @@
...
+        adapter = communicator.createObjectAdapterWithEndpoints("GreeterAdapter", "tcp -p 4061")
+
+        # Register the Chatbot servant with the adapter.
+        adapter.add(chatbot.Chatbot(), Ice.Util.stringToIdentity("greeter"))
+
+        # Start dispatching requests.
+        adapter.activate()
+        print("Listening on port 4061...")
+
+        try:
```
bernardnormier, line 39: Use the simpler model here.

bernardnormier (approved):

python/Ice/greeter/server_asyncio/main.py
pepone: I don't think we can call waitForShutdown here directly, as this is the event loop thread, and would block it. My understanding is that we would need an `Ice.Future` version and then wrap the future to use it in the asyncio loop.

python/Ice/greeter/server_asyncio/main.py
externl: Maybe we should also say what this coroutine executor is used for.

externl (approved):

PR #322, opened 2025-02-26
New IceGrid/Greeter demo (C#)

bernardnormier:
This PR adds a new IceGrid/Greeter demo in C#.

It's similar to the existing simple demo except it does not show Query or LocatorDiscovery.  I plan to add separate Query and LocatorDiscovery demos in a follow-up PR.


csharp/IceGrid/Greeter/application-with-replication.xml
```diff
@@ -0,0 +1,29 @@
+<!-- Copyright (c) ZeroC, Inc. All rights reserved. -->
```
pepone, line 1:
```suggestion
<!-- Copyright (c) ZeroC, Inc. -->
```

csharp/IceGrid/Greeter/application-with-replication.xml
```diff
@@ -0,0 +1,29 @@
...
+
+<icegrid>
+
+  <application name="VisitorCenter">
+
+    <server-template id="GreetServer">
+      <parameter name="index"/>
+      <server id="GreetServer-${index}" exe="dotnet" activation="on-demand">
+        <option>Server/bin/Debug/net8.0/Server.dll</option>
```
pepone, line 10: hardcoding Debug here would not work when user build the Release configuration.

csharp/IceGrid/Greeter/application-with-replication.xml
```diff
@@ -0,0 +1,29 @@
+<!-- Copyright (c) ZeroC, Inc. All rights reserved. -->
+
```
pepone, line 2: Would be nice to have comments explaining the deployment in this file.

csharp/IceGrid/Greeter/application.xml
```diff
@@ -0,0 +1,19 @@
+<!-- Copyright (c) ZeroC, Inc. All rights reserved. -->
```
pepone, line 1:
```suggestion
<!-- Copyright (c) ZeroC, Inc. -->
```

csharp/IceGrid/Greeter/application.xml
```diff
@@ -0,0 +1,19 @@
...
+
+<icegrid>
+
+  <application name="VisitorCenter">
+
+    <node name="node1">
+
+      <server id="GreetServer" exe="dotnet" activation="on-demand">
+        <option>Server/bin/Debug/net8.0/Server.dll</option>
```
pepone, line 10: ditto for `Debug/`

externl:
Maybe we should do an extra publish step:
eg.

```
dotnet publish Server -o bin/
```

csharp/IceGrid/Greeter/application-with-replication.xml
bernardnormier: Yes, will do in a follow-up PR.

bernardnormier:
> Maybe we should do an extra publish step: eg.
> 
> ```
> dotnet publish Server -o bin/
> ```
> 
> Then use `bin/Server` in IceGrid. This is likely more in line with a real deployment.

I would rather keep all Server-related files in the Server subdir, and there is already Server/bin with Debug/net8.0/...

externl (approved):

csharp/IceGrid/Greeter/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Follow these steps to build and run the demo:
+
+1. Build the client and server applications:
+
+```shell
+dotnet build
+```
+
+2. Start the IceGrid registry in its own terminal:
```
bernardnormier, line 13: Need instructions on obtaining/installing icegridregistry & icegridnode.

csharp/IceGrid/Greeter/application-with-replication.xml
pepone: If we want to keep this for simplicity we should mention it in the README.

pepone (approved):

PR #323, opened 2025-02-27
Improve IceGrid/Greeter

bernardnormier:
This PR improves IceGrid/Greeter in C#, with in particular comments in the XML files.

### DIFF
--- a/csharp/IceGrid/Greeter/Server/Program.cs
+++ b/csharp/IceGrid/Greeter/Server/Program.cs
@@ -20,14 +20,5 @@ adapter.add(new Server.Chatbot(greeterName), greeterIdentity);
 adapter.activate();
 Console.WriteLine($"{greeterName} is listening...");
 
-// Shut down the communicator the program receives a SIGTERM or SIGINT signal.
-Console.CancelKeyPress += (sender, eventArgs) =>
-{
-    eventArgs.Cancel = true; // don't terminate the process
-    Console.WriteLine("Caught signal, shutting down...");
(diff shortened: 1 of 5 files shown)


csharp/IceGrid/Greeter/Server/Program.cs
```diff
@@ -20,14 +20,5 @@
 adapter.activate();
 Console.WriteLine($"{greeterName} is listening...");
 
-// Shut down the communicator the program receives a SIGTERM or SIGINT signal.
-Console.CancelKeyPress += (sender, eventArgs) =>
```
bernardnormier, line 24: This is not used at all when the server registers its Process facet, which is always the case.

csharp/IceGrid/Greeter/application-with-replication.xml
```diff
@@ -1,29 +1,41 @@
...
+    <!-- Represents a set of replicated object adapters. -->
+    <replica-group id="GreeterPool">
+      <!-- Specifies in which order IceGrid returns the object adapters in the pool.-->
       <load-balancing type="round-robin"/>
+
+      <!-- Defines the well-known object "greeter", implemented by all replicas. -->
       <object identity="greeter"/>
     </replica-group>
 
+    <!-- Describes the servers deployed on node1. Here, we deploy 3 separate instances of the GreetServer template. -->
```
pepone, line 34:
```suggestion
    <!-- Describes the servers deployed on node1. Here, we deploy 3 separate instances of the GreetServer. -->
```

pepone (approved):

PR #324, opened 2025-02-27
Update Filesystem client to make async invocations (C#)

### LINKED ISSUE #308: Upgrade  Ice/Filesystem to use async invocations
The C#  Ice/Filesystem demo should be updated to use async invocations, in general C# demos should use async/await for RPCs.

bernardnormier:
Fixes #308.

### DIFF
--- a/csharp/Ice/Filesystem/Client/Program.cs
+++ b/csharp/Ice/Filesystem/Client/Program.cs
@@ -28,11 +28,11 @@ async Task ListRecursiveAsync(DirectoryPrx dir, int depth = 0)
     {
         Debug.Assert(node is not null); // The node proxies returned by list() are never null.
 
-        // TODO: should be an async call
-        DirectoryPrx? subdir = DirectoryPrxHelper.checkedCast(node);
+        DirectoryPrx? subdir = await DirectoryPrxHelper.checkedCastAsync(node);
         string kind = subdir is not null ? "directory" : "file";
+        string nodeName = await node.NameAsync();
 
-        Console.WriteLine($"{indent}{await node.NameAsync()} {kind}:");
+        Console.WriteLine($"{indent}{nodeName} {kind}:");
 
         if (subdir is not null)
         {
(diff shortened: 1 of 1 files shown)


csharp/Ice/Filesystem/Client/Program.cs
```diff
@@ -12,35 +12,36 @@
...
     var indent = new string('\t', ++depth);
 
-    NodePrx?[] contents = dir.List();
+    NodePrx?[] contents = await dir.ListAsync();
 
     foreach (NodePrx? node in contents)
     {
         Debug.Assert(node is not null); // The node proxies returned by list() are never null.
 
+        // TODO: should be an async call
```
bernardnormier, line 31: I think we should generate a checkedCastAsync ... anything else would be inconsistent.

pepone: Adding checkedCastAsync sounds good. On a side note the Slice model for this example isn't great.

I wondering if we should add a NodeInfo struct  and NodeKind enum:

```
enum NodeKind
{
   File,
   Directory
}

struct NodeInfo
{
    string name;
    NodeKind kind;
}

sequence<NodeInfo> NodeInfoSeq;

interface Directory extends Node
{
    idempotent NodeInfoSeq list();
}
```

This avoids the need for separate name, checkedCast requests.

csharp/Ice/Filesystem/Client/Program.cs
```diff
@@ -12,35 +12,36 @@
...
         DirectoryPrx? subdir = DirectoryPrxHelper.checkedCast(node);
         string kind = subdir is not null ? "directory" : "file";
 
-        Console.WriteLine($"{indent}{node.Name()} {kind}:");
+        Console.WriteLine($"{indent}{await node.NameAsync()} {kind}:");
```
pepone, line 35: I would add a name variable here.

pepone (approved):

csharp/Ice/Filesystem/Client/Program.cs
bernardnormier: This demo is really about the power of inheritance. And implementation reuse in C++. So I would not change the model or the `checkedCast[Async]`.

PR #326, opened 2025-02-28
Improvements to CMake builds on Windows

externl:
This PR adds several improvements for the CMake builds on Windows.

- Add support for [TARGET_RUNTIME_DLLS](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_RUNTIME_DLLS) generator expressions. This required some restructuring to tell cmake about the dlls and the libs. Now we can configure cmake to copy a target dependencies next to the target post build. No need for users to add a random NuGet dir to their PATH.
- The unzipped Ice nuget is pretty big. Instead of a per config "cmake build folder" copy of the nuget, we now keep (git ignored) a shared one in `cpp/cmake/packages`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,8 +36,5 @@
         "${workspaceFolder}/cpp/IceStorm/counter",
         "${workspaceFolder}/cpp/IceStorm/replicated",
         "${workspaceFolder}/cpp/IceStorm/replicated2",
-    ],
-    "cmake.configureSettings": {
-        "Ice_HOME": "${workspaceFolder}/../ice",
-    }
+    ]
 }
--- a/cpp/Ice/greeter/CMakeLists.txt
+++ b/cpp/Ice/greeter/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.21)
 
 project(greeter CXX)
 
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
externl, line 1: We no longer need this!

cpp/Ice/greeter/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.21)
```
externl, line 1: `TARGET_RUNTIME_DLLS` is a 3.21 feature.

cpp/cmake/IceConfig.cmake
```diff
@@ -24,10 +27,10 @@ endif()
 if(Ice_SOURCE_BUILD)
```
pepone, line 27: For other languages we are no longer supporting to use the source build.

cpp/cmake/common.cmake
```diff
@@ -1,52 +1,58 @@
-cmake_minimum_required(VERSION 3.24)
+cmake_minimum_required(VERSION 3.21)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+if (WIN32)
+  set(Ice_NUGET_NAME "zeroc.ice.v143")
+  set(Ice_NUGET_DIR "${CMAKE_CURRENT_LIST_DIR}/packages/${Ice_NUGET_NAME}")
```
pepone, line 8: I guessing that I can use a custom NuGet build by just installing it to this location, which would skip the install below.

cpp/cmake/IceConfig.cmake
```diff
@@ -24,10 +27,10 @@ endif()
...
   list(APPEND ice_include_path_suffixes "cpp/include")
   list(APPEND ice_bin_path_suffixes "cpp/bin")
-  list(APPEND ice_lib_path_suffixes_release "cpp/lib")
+  list(APPEND ice_lib_path_suffixes "cpp/lib")
 
   if (Ice_ARCHITECTURE)
```
pepone, line 32: What is Ice_ARCHITECTURE and when is it not-set?

pepone (approved):

cpp/cmake/IceConfig.cmake
externl: It's computed from a few values (see above).

- If `CMAKE_LIBRARY_ARCHITECTURE` is set we use it. For example this is set on Ubuntu to something like "x86_64-linux-gnu"
- Otherwise if not set and we're on Windows we'll set it to either x64 or Win32

If it's not set nothing happens, it's a matter of checking `lib` vs `lib/<arch>`

PR #332, opened 2025-03-03
Update java/Ice/greeter shutdown

bernardnormier:
This PR update the shutdown of the java/Ice/greeter demo to be more like the shutdown of demos in other languages.

### DIFF
--- a/java/Ice/greeter/README.md
+++ b/java/Ice/greeter/README.md
@@ -39,7 +39,7 @@ First, start one of the server implementations:
 ### Windows
 
 ```shell
-server\build\install\server\binserver
+server\build\install\server\bin\server
 ```
 
 or
--- a/java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
+++ b/java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
@@ -2,17 +2,18 @@
 
 package com.zeroc.demos.greeter;
 
-import com.zeroc.demos.greeter.Chatbot;
 import com.zeroc.Ice.Communicator;
 import com.zeroc.Ice.ObjectAdapter;
 
 class Server {
     public static void main(String[] args) {
-        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+        // Create an Ice communicator to initialize the Ice runtime. The communicator is closed (destroyed) at the end
+        // of the try statement.
         try (Communicator communicator = com.zeroc.Ice.Util.initialize(args)) {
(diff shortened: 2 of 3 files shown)


java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
```diff
@@ -2,17 +2,18 @@
 
 package com.zeroc.demos.greeter;
 
-import com.zeroc.demos.greeter.Chatbot;
```
bernardnormier, line 5: Don't need import since Chatbot is in the same package.

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
```diff
@@ -24,9 +25,22 @@ public static void main(String[] args) {
...
             System.out.println("Listening on port 4061...");
 
-            // Wait until the user presses Ctrl+C.
+            // Wait until the communicator is shut down. Here, this occurs when the user presses Ctrl+C.
             communicator.waitForShutdown();
-            System.out.println("Caught Ctrl+C, exiting...");
         }
     }
+
+    private static void shutdownCommunicatorOnCtrlC(Communicator communicator, Thread cleanupThread) {
```
pepone, line 33: I don't like cleanupThread name very much, what about using just mainThread?

pepone (approved):

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
externl: I also prefer mainThread. It's not obvious from the name cleanupThread that your intention is that a user would do cleanup in the main thread, since we don't do any. Especially since the contention in Java seems to be that you would do the cleanup in this hook.

externl (approved):

InsertCreativityHere (approved):

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Server.java
externl: Like Jose suggested, I would just call it mainThread.

java/Ice/greeter/README.md
```diff
@@ -39,7 +39,7 @@ First, start one of the server implementations:
 ### Windows
 
 ```shell
-server\build\install\server\binserver
```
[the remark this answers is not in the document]
pepone: I think we change this because how Ctrl-C is handled by the scripts.

PR #333, opened 2025-03-04
Polishing the 'Java/greeter' Demo

InsertCreativityHere:
This PR puts some final polish on the Java/greeter demo.

The biggest change is to the instructions, where we now recommend
`./gradlew :server:run --quiet` instead of `server\build\install\server\bin\server`
to run the generated demo executables.

----

I also removed the distinction between `mac/linux` and `windows`. The only difference was whether a `./` was in front.
Nowadays, the _default_ terminal on Windows is PowerShell, which _requires_ a leading `./` in front to work. Same as mac/linux.
EDIT: I should say that Powershell also accepts a leading `.\`. It's generally tolerant of either kind of slash.

The commands we were recommending were clearly intended for `cmd` users.
But, `cmd` is 'old' now. We should be focusing on PowerShell. And this makes the README much shorter and easier to follow.
For the users still attached to `cmd`, I'm sure they can figure out to omit the leading `./` on their own.


java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -9,7 +9,7 @@
 
 class Client {
     public static void main(String[] args) {
-        // Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program
```
InsertCreativityHere, line 12: `disposed` in a C# specific term; In Java, objects are _destroyed_.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -22,17 +22,16 @@ public static void main(String[] args) {
             String greeting = greeter.greet(System.getProperty("user.name"));
             System.out.println(greeting);
 
-            // Send another request to the remote object, this time with greetAsync. greetAsync returns a future
```
InsertCreativityHere, line 25: Felt weird to have a sentence start with a lower-case letter...

java/Ice/greeter/server/src/main/java/com/zeroc/demos/greeter/Chatbot.java
```diff
@@ -10,7 +10,7 @@
  */
 class Chatbot implements Greeter
 {
-    // Implements the abstract method Greet from the Greeter interface generated by the Slice compiler.
```
InsertCreativityHere, line 13: Should be camelCase in Java.

java/Ice/greeter/serveramd/src/main/java/com/zeroc/demos/greeter/Chatbot.java
```diff
@@ -13,9 +13,9 @@
  */
 class Chatbot implements Greeter
 {
-    // Implements the abstract method Greet from the Greeter interface generated by the Slice compiler.
-    // This variant is the asynchronous implementation.
-    public CompletionStage<String>  greetAsync(String name, Current current) {
```
InsertCreativityHere, line 18: It felt weird to say a method named `greetAsync` implements the `greet` method, without a little explanation.

java/Ice/greeter/README.md
```diff
@@ -12,64 +12,32 @@ The same client works with both.
...
-
-```shell
-./server/build/install/server/bin/server
-```
-
-### Windows
-
 ```shell
-server\build\install\server\bin\server
+./gradlew :server:run --quiet
```
bernardnormier, line 26: The title should be 'Running the Server'. You should only run one server at a time.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
bernardnormier: See the comment in Server.java

java/Ice/greeter/serveramd/src/main/java/com/zeroc/demos/greeter/Chatbot.java
bernardnormier: the method name is obviously greetAsync.

java/Ice/greeter/client/src/main/java/com/zeroc/demos/greeter/Client.java
InsertCreativityHere: I updated the comment to be the same!
I thought the difference in wording was intentional...

java/Ice/greeter/serveramd/src/main/java/com/zeroc/demos/greeter/Chatbot.java
InsertCreativityHere: Fixed!
And yes I agree, I mistakenly thought there was some other reason for calling it `greet`... so I kept it.
But yeah, it should definitely be `greetAsync`.

java/Ice/greeter/README.md
InsertCreativityHere: Fair enough, I made it singular.

java/Ice/greeter/README.md
```diff
@@ -12,64 +12,32 @@ The same client works with both.
...
-In a separate terminal, start the client program:
-
-### Linux/macOS
-
-```shell
-./client/build/install/client/bin/client
-```
-
-### Windows
+Then, in a separate terminal, start the client program:
```
bernardnormier, line 37: We should program call it client application and server application. Program comes from the C# demo.

bernardnormier (approved):

java/Ice/greeter/README.md
InsertCreativityHere: Updated it to say "application".

pepone (approved):

PR #336, opened 2025-03-04
CMake config file fixes

externl:
This PR makes several Ice CMake fixes in preparation to move the cmake files into the Ice packaging. It also adds a dev container.

- Split into multiple CMake files.  A general config that loads the targets and the slice compiler function.
- Target loading fixes to explicitly set dependencies and c++ version


cpp/cmake/IceConfig.cmake
```diff
@@ -1,239 +1,16 @@
...
-endif()
+# get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)
 
-include(CMakeFindDependencyMacro)
-find_package(Threads REQUIRED)
-
-if(NOT DEFINED Ice_ARCHITECTURE)
-  if (CMAKE_LIBRARY_ARCHITECTURE)
-    set(Ice_ARCHITECTURE ${CMAKE_LIBRARY_ARCHITECTURE} CACHE STRING "Library architecture")
+if(NOT DEFINED PACKAGE_PREFIX_DIR)
```
externl, line 5: These lines are temporary and will be replace with the above line once we move into the Ice packaging.

This is the standard setup for cmake as `IceConfig.cmake` is placed at `<prefix>/lib/cmake/ice/IceConfig.cmake`

cpp/cmake/slice2cpp.cmake
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) ZeroC, Inc.
+
+find_program(Ice_SLICE2CPP_EXECUTABLE slice2cpp
+  HINTS ${PACKAGE_PREFIX_DIR}
+  PATH_SUFFIXES bin ../../tools
```
externl, line 5: I think this relative path is correct for the nuget (CI will tell us). Would be nice to find a better way for this.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+USER vscode
+WORKDIR /home/vscode
+
+RUN set -eux \
+    && wget -q https://packages.microsoft.com/config/ubuntu/24.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
+    && sudo dpkg -i packages-microsoft-prod.deb \
+    && rm packages-microsoft-prod.deb \
+    && sudo apt update \
+    && sudo apt-get install -y php python3 python3-passlib ruby-full gdb \
```
pepone, line 11: Do we need passlib for the demos?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+    // Use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+    "build": {
+        "dockerfile": "Dockerfile"
+    },
+    "runArgs": [
```
pepone, line 6: What are the defaults when this is commented out?

cpp/cmake/IceTargets.cmake
```diff
@@ -0,0 +1,158 @@
...
+# Add an imported target for slice2cpp
+add_executable(Ice::slice2cpp IMPORTED)
+set_target_properties(Ice::slice2cpp PROPERTIES
+  IMPORTED_LOCATION "${Ice_SLICE2CPP_EXECUTABLE}"
+)
+
+find_path(Ice_SLICE_DIR
+  NAMES Ice/Identity.ice
+  HINTS ${PACKAGE_PREFIX_DIR}
+  PATH_SUFFIXES slice share/ice/slice slice
```
pepone, line 48: Do we need `slice` twice here?

cpp/cmake/slice2cpp.cmake
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) ZeroC, Inc.
+
+# Function to generate C++ source files from Slice (.ice) files for a target using slice2cpp
+# The target must have the Slice files in its sources
+# The generated files are added to the target sources
+# Usage:
+# add_executable(a_target source1.cpp source2.ice source3.ice)
+# slice2cpp_generate(a_target)
+function(slice2cpp_generate target)
```
pepone, line 9: Did you think on how to handle Slice file dependencies to avoid rebuilds?

pepone (approved):

.devcontainer/Dockerfile
externl: No, I'll remove it!

.devcontainer/devcontainer.json
externl: Just the default like when you run a regular docker container.

cpp/cmake/IceTargets.cmake
externl: No. It's a typo.

cpp/cmake/slice2cpp.cmake
externl: No, I need to look into this.

PR #337, opened 2025-03-04
New Glacier2/Callback demo

bernardnormier:
This PR adds a new Glacier/Callback demo in C#.


csharp/Glacier2/Callback/Client/Client.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <OutputType>Exe</OutputType>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <!-- Copy the PDBs from the NuGet packages to get file names and line numbers in stack traces. -->
+    <CopyDebugSymbolFilesFromPackages>true</CopyDebugSymbolFilesFromPackages>
+  </PropertyGroup>
+  <ItemGroup>
+    <SliceCompile Include="../slice/AlarmClock.ice" />
+    <PackageReference Include="ZeroC.Ice" Version="$(IceVersion)" />
+    <PackageReference Include="ZeroC.Glacier2" Version="$(IceVersion)" />
```
pepone, line 14: We are explicitly adding all dependencies, but should we instead rely on transitive dependencies? For example, adding Glacier2 should automatically include Ice.

pepone (approved):
Looks good, I was thinking if we can just reuse the Bidir server, but I guess the setup would be a bit more complicated.

csharp/Glacier2/Callback/Client/Client.csproj
externl: I'm not sure what the typical thing to do in C# is, but since the client uses Ice I don't find it to bad to have it listed.

externl (approved):

csharp/Glacier2/Callback/Client/Client.csproj
bernardnormier: I think it's ok to reference Ice and not use transitive dependency here.

In the IceRPC Slice examples, we let `IceRpc.Slice` pull `IceRpc`, but I feel it's a bit different here:
 - IceRpc.Slice sounds like IceRpc + Slice (what you are using in the example)
 - ZeroC.Glacier2 does not sound like Ice + Glacier2 ... it's just Glacier2.

bernardnormier:
> Looks good, I was thinking if we can just reuse the Bidir server, but I guess the setup would be a bit more complicated.

We can't reuse the Bidir server: the Bidir server needs to be aware of the client.

However, as described in the README, the server is identical to the Ice/Callback server. Rather than ask the user to start the Ice/Callback server, I copied it into this new demo.

InsertCreativityHere (approved):
Looks good to me!

PR #338, opened 2025-03-05
New IceGrid/IceBox demo in C#

bernardnormier:


csharp/IceGrid/IceBox/simpsons.xml
```diff
@@ -0,0 +1,42 @@
...
+            <property name="Greeter.Name" value="${name}" />
+         </service>
+    </service-template>
+
+    <!-- Describes the servers deployed on the IceGrid node named 'node1'. -->
+    <node name="node1">
+
+      <!-- Describes the server deployed on node1. IceGrid starts this server when a client looks up an object adapter
+      or a well-known object in this server. Here, we deploy a single IceBox server that loads 4 IceBox services. -->
+      <icebox id="IceBox" activation="on-demand" exe="dotnet">
```
bernardnormier, line 32: The previous demo (old-icebox) deployed the services as replicas, which I don't think is appropriate here. Each service produces a distinct greeting ... they are not replicas.

csharp/IceGrid/IceBox/config.node
pepone, line 1: Our naming convention for this config files is odd. I think `node.config`, `admin.config` makes more sense. Something to think for a future PR.

bernardnormier: I agree, and we use *.cfg for the model config files in ice/cpp/config:
```
ls *.cfg
glacier2router.cfg	icegridnode.cfg		icegridregistry.cfg
```

Should we update all demo config files to use *.cfg?

What about the config files generated by icegridnode?
```
ls servers/IceBox/config
config		config_Bart	config_Homer	config_Lisa	config_Maggie	config_Marge
```

Keep as is?

csharp/IceGrid/IceBox/Client/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+using Ice.Communicator communicator = Ice.Util.initialize(ref args);
+
+// Set the default locator of the new communicator. It's the address of the Locator hosted by our IceGrid registry.
+// You can also set this proxy with the Ice.Default.Locator property.
+communicator.setDefaultLocator(
+    Ice.LocatorPrxHelper.createProxy(communicator, "IceGrid/Locator:tcp -h localhost -p 4061"));
+
+// Create a proxy to the Query object hosted by the IceGrid registry. "IceGrid/Query" a well-known proxy, without
+// addressing information.
+IceGrid.QueryPrx query = IceGrid.QueryPrxHelper.createProxy(communicator, "IceGrid/Query");
```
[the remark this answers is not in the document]
pepone: Yes simpler would be better.  Do we really need this demo?

- We have IceGrid/Greeter that show how to deploy a server, including a replicated setup
- We have IceBox/Greeter that shows how to create an IceBox service.

There isn't much new here or specific to IceBox.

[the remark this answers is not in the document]
bernardnormier: There are a few things specific to IceBox - the XML support for `<icebox>`, with IceBox services described in XML, one config file generated per config and more.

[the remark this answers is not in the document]
bernardnormier: Simplified demo, with no Query or template.

csharp/IceGrid/IceBox/config.node
externl: I also find our naming convention odd and would prefer to switch to `.config`. 

Not a big fan of `.cfg`. I'm pretty sure we actually install them as `glacier2config.conf`, etc. 

I'm not opposed to renaming the ones generated by `icegridnode` but I don't think it's as important.

externl (approved):

pepone (approved):

PR #339, opened 2025-03-05
Add IceDiscovery/Replication demo in C#

bernardnormier:


csharp/IceDiscovery/Replication/Client/Program.cs
```diff
@@ -0,0 +1,25 @@
...
+initData.properties.setProperty("Ice.Plugin.Discovery", "IceDiscovery:IceDiscovery.PluginFactory");
+
+// Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+using Ice.Communicator communicator = Ice.Util.initialize(initData);
+
+// Create proxy to the Greeter object hosted by the server. "greeter" is a stringified proxy with no addressing
+// information, also known as a well-known proxy. It's resolved by the default locator installed by the IceDiscovery
+// plugin.
+GreeterPrx greeter = GreeterPrxHelper.createProxy(communicator, "greeter");
+
```
pepone, line 21: One issue with these IceDiscovery demos is that we assume the locator was set at this point. If I recall correctly the plug-in does this in the background, and there is no API to wait for it.

pepone (approved):

csharp/IceDiscovery/Replication/Client/Program.cs
externl: Doesn't it do in during plugin init?

externl (approved):

csharp/IceDiscovery/Replication/Client/Program.cs
bernardnormier: I am not aware of any issue or bug with the IceDiscovery plugin initialization.

PR #341, opened 2025-03-06
Remove old C# demos

bernardnormier:
This PR removes a number of old C# demos.


pepone (approved):
Looks good, we need to upgrade the READMEs.

externl (approved):

PR #350, opened 2025-03-09
Turn on AnalysisMode=all in C#

bernardnormier:


.editorconfig
```diff
@@ -63,6 +63,9 @@ dotnet_style_readonly_field = true:suggestion
 # Parameter preferences
 dotnet_code_quality_unused_parameters = all:suggestion
 
+# Diagnostic preferences
+dotnet_diagnostic.CA5394.severity = none # Do not use insecure randomness
```
bernardnormier, line 67: Otherwise, can't use the pseudo-random generator System.Random.

csharp/Directory.Build.props
```diff
@@ -3,11 +3,7 @@
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
         <IceVersion Condition="'$(IceVersion)' == ''">3.8.0-nightly.*</IceVersion>
-        <ImplicitUsings>true</ImplicitUsings>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <!-- <AnalysisMode>All</AnalysisMode> -->
-        <ImplicitUsings>enable</ImplicitUsings>
```
bernardnormier, line 9: This is / should be specified in all .csproj files.

pepone (approved):

.editorconfig
externl: Instead we should be using ` RandomNumberGenerator.GetInt32` like we do in IceRPC.

externl (approved):

.editorconfig
bernardnormier: Will fix in a follow-up PR.

PR #352, opened 2025-03-10
common.cmake cleanup

externl:


pepone (approved):
Looks good but why do we keep all this extra logic to install NuGet packages. We don't do the same with other distributions.

Wouldn't it be simpler to just document `nuget install` usage.

externl:
I got the original idea from other cmake example I saw (though I don't remember where). I think it's a nice way to show how you can integrate the NuGet package into CMake as part of the build system. If other distributions had  non-global installs I would be included to show how to add them with cmake as well.

PR #356, opened 2025-03-17
Use shutdownCompleted to wait for communicator shutdown.

pepone:
Update asyncio greeter server to use shutdownCompleted.

### DIFF
--- a/python/Ice/greeter/server_asyncio/main.py
+++ b/python/Ice/greeter/server_asyncio/main.py
@@ -26,7 +26,7 @@ async def main():
     # Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the with block.
     with Ice.initialize(initData) as communicator:
 
-        # Shutdown the communicator when the user presses Ctrl-C.
+        # Shutdown the communicator when the user presses Ctrl+C.
         loop.add_signal_handler(signal.SIGINT, communicator.shutdown)
 
         # Create an object adapter that listens for incoming requests and dispatches them to servants.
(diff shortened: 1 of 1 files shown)


python/Ice/greeter/server_asyncio/main.py
```diff
@@ -24,6 +25,10 @@ def coroutineExecutor(coroutine):
 
     # Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the with block.
     with Ice.initialize(initData) as communicator:
+
+        # Shutdown the communicator when the user presses Ctrl-C.
```
bernardnormier, line 29: We use Ctrl+C in other demos and below.

python/Ice/greeter/server_asyncio/main.py
```diff
@@ -34,11 +39,7 @@ def coroutineExecutor(coroutine):
         adapter.activate()
         print("Listening on port 4061...")
 
-        try:
-            # Wait until the event completes, which only occurs when it is canceled by Ctrl+C.
-            exit_event = asyncio.Event()
-            await exit_event.wait()
-        except asyncio.exceptions.CancelledError:
-            print("Caught Ctrl+C, exiting...")
+        # Wait until the communicator is shut down. Here, this occurs when the user presses Ctrl+C.
```
bernardnormier, line 42: We should add a comment about `Ice.wrap_future`, as it's not obvious. I assume it means we create an asyncio-compatible future from an Ice fture?

bernardnormier (approved):

python/Ice/greeter/server_asyncio/main.py
pepone: Not clear why we have Ice.wrap_future, when there is asyncio.wrap_future for doing this.

https://docs.python.org/3/library/asyncio-future.html#asyncio.wrap_future

externl (approved):

PR #360, opened 2025-03-27
Add JavaScript Glacier2 callback demo

pepone:
Ports csharp/Glacier2/Callback demo to JavaScript.


js/Glacier2/callback/MockAlarmClock.ts
externl, line 1: Missing copyright.

js/Glacier2/callback/timeUtil.ts
```diff
@@ -0,0 +1,8 @@
+const millisecondsBeforeEpoch = 719162n * 24n * 60n * 60n * 1000n; // daysBeforeEpoch converted to milliseconds
```
externl, line 1: Should we make put this in a common top level folder for use by future demos?

externl (approved):

js/Glacier2/callback/timeUtil.ts
```diff
@@ -0,0 +1,8 @@
...
+const ticksPerMillisecond = 10000n; // number of ticks per millisecond
+
+// Convert a timestamp representing the number of milliseconds elapsed since the epoch to a BigInt representing
```
bernardnormier, line 4: Should this be a doc-comment?

We should also say "Unix epoch". There is not one universal epoch. In the C++ code, we use system_clock's epoch, which is the Unix epoch (since C++20, but hopefully before as well).
See https://en.cppreference.com/w/cpp/chrono/system_clock

js/Glacier2/callback/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "lib": ["dom", "es2020"],
+        "target": "es2020",
+        "module": "es2022", // "es2022" required top-level 'await using' statements
```
bernardnormier, line 5: required for top-level await using?

js/Glacier2/callback/MockAlarmClock.ts
```diff
@@ -0,0 +1,23 @@
+import { Ice } from "@zeroc/ice";
+import { EarlyRiser } from "./AlarmClock.js";
+
+export class MockAlarmClock extends EarlyRiser.AlarmClock {
+    private _needMoreTime = true;
+    private _resolver: (value: void | PromiseLike<void>) => void;
+
+    public StopPressed = new Promise<void>((resolve, reject) => (this._resolver = resolve));
```
bernardnormier, line 8: Is StopPressed created when MockAlarmClock is created or the first time StopPressed is used?

js/Glacier2/callback/client.ts
```diff
@@ -0,0 +1,56 @@
...
+// invoked through the Glacier2 router.
+const clientCategory = await router.getCategoryForClient();
+
+// Create an object adapter with no name and no configuration, but with our router proxy. This object adapter is a
+// "bidirectional" object adapter, like the one created by the Ice/bidir client application. It does not listen on any
+// port and it does not need to be activated.
+const adapter = await communicator.createObjectAdapterWithRouter("", router);
+
+// Register the MockAlarmClock servant with the adapter. It uses the category retrieved from the router. You can verify
+// the Ring callback is never delivered if you provide a different category.
```
bernardnormier, line 37: lower-case ring

js/Glacier2/callback/README.md
```diff
@@ -0,0 +1,42 @@
...
+The connection between the client and the Glacier2 router is a "bidir" connection, like in the [Ice Bidir][2] demo:
+
+```mermaid
+flowchart LR
+    c[Client<br>hosts AlarmClock] --bidir connection--> g[Glacier2 router:4063]
+    g --connection1--> s[Server:4061<br>hosts WakeUpService] --connection2--> g
+```
+
+Ice for JavaScript has limited server-side support ([see documentation][1]). As a result, you need to start a Glacier2
+Callback server implemented in a language that fully supports server-side functionality, such as Python, Java, or C#.
```
bernardnormier, line 15: You also need to start a Glacier2 router.

bernardnormier (approved):

js/Glacier2/callback/timeUtil.ts
pepone: We already use it in a few demos. I will refactor this in a follow up PR.

js/Glacier2/callback/MockAlarmClock.ts
pepone: When it is created.

The generated JavaScript looks like:

```
export class MockAlarmClock extends EarlyRiser.AlarmClock {
    constructor() {
        super(...arguments);
        this._needMoreTime = true;
        this.StopPressed = new Promise((resolve, reject) => (this._resolver = resolve));
    }
```

PR #361, opened 2025-03-28
Add JavaScript Ice/cancellation demo

pepone:
This is a port of the C# Cancellation demo to JavaScript. Client side only.


js/Ice/cancellation/client.ts
```diff
@@ -0,0 +1,61 @@
...
+// Retrieve my name
+const name = process.env.USER || process.env.USERNAME || "masked user";
+
+// Send a request to the regular greeter and get the response.
+let greeting = await greeter.greet(name);
+console.log(greeting);
+
+// Create another slow greeter proxy with an invocation timeout of 4 seconds (the default invocation timeout is
+// infinite).
+const slowGreeter4s = slowGreeter.ice_invocationTimeout(4000);
```
externl, line 28: Should we move this next to the other geeters?

externl (approved):

js/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+|                                           |                                                                                                     |
+|-------------------------------------------|-----------------------------------------------------------------------------------------------------|
+| [Glacier2 Callback](./Glacier2/callback/) | Shows how to write a Glacier2 client and implement callbacks in this client.                        |
+| [Ice Bidir](./Ice/bidir/)                 | Shows how to send requests "the other way around", from the server to the client.                   |
+| [Ice Cancellation](./Ice/cancellation/)   | Shows how to cancel a request and how to configure invocation timeouts.                             |
+| [Ice Context](./Ice/context/)             | Shows how to set and retrieve request contexts.                                                     |
+| [Ice Filesystem](./Ice/filesystem/)       | Shows the power of interface inheritance in Slice.                                                  |
+| [Ice Greeter](./Ice/greeter/)             | Shows how to call and implement a canonical Greeter application with Ice. **Start with this demo!** |
```
externl, line 28: We removed this list for the other languages.

externl (approved):

js/Ice/cancellation/client.ts
bernardnormier: Yes, and we should fix C# too.

js/Ice/cancellation/client.ts
```diff
@@ -0,0 +1,61 @@
...
+} catch (exception) {
+    console.assert(exception instanceof Ice.InvocationCanceledException, exception);
+    console.log(`Caught InvocationCanceledException, as expected: ${exception.message}`);
+}
+
+// Verify the regular greeter still works.
+greeting = await greeter.greet("carol");
+console.log(greeting);
+
+// Send a request to the slow greeter, and wait forever for the response.
```
bernardnormier, line 54: Not sure about this last invocation:

- it's about server-side dispatch cancellation/interruption, which in this demo doesn't use JS
- a JS app can cancel a bidir dispatch, which would throw the same exception to the caller... but it's a different setup

Should we remove this last part?

js/README.md
bernardnormier: That's the new model that we use in C#. Keep it!

bernardnormier (approved):
Looks good.

PR #363, opened 2025-03-28
Add Python bidir demo

pepone:
This is a port of the  new C# Ice/bidir demo to Python.


python/Ice/bidir/README.md
```diff
@@ -1,18 +1,76 @@
...
 
+```shell
+pip install -r requirements.txt
 ```
-python client.py
+
+- Compile the AlarmClock.ice file with the Slice compiler for Python:
+
+```shell
+slice2py ../slice/AlarmClock.ice
```
bernardnormier, line 37: slice2py gets installed by `pip install` above?

python/common/common/time.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) ZeroC, Inc.
+
+"""
+Utility functions to convert between datetime objects and custom timestamps
```
bernardnormier, line 4: One difficulty here is Python uses the term timestamp already, with a different meaning:
https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp

And, as far as can tell, it doesn't use the term time point.

Then, it appears that the Jan 1, 0001 model we use - the proleptic Gregorian calendar - is actually the default for Python. So we probably don't need these functions at all.

python/Ice/bidir/server/server.py
```diff
@@ -0,0 +1,44 @@
...
+# Copyright (c) ZeroC, Inc.
+
+import asyncio
+import time
+import Ice
+from EarlyRiser import ButtonPressed, WakeUpService, AlarmClockPrx
+from common.time import toTimePoint
+
+
+class BidirWakeUpService(WakeUpService):
```
bernardnormier, line 11: Missing comments

python/Ice/bidir/server/server.py
```diff
@@ -0,0 +1,44 @@
...
+        connection = current.con # The connection from the client to the server.
+        if (connection is None):
+            # Should never happen, but in case it does, the Ice runtime transmits this exception as an
+            # Ice.UnknownException.
+            raise NotImplementedError("BidirWakeUpService does not support collocated calls")
+        
+        # Create a proxy to the client's alarm clock. This connection-bound proxy is called a "fixed proxy".
+        alarmClock = AlarmClockPrx.uncheckedCast(connection.createProxy(Ice.stringToIdentity("alarmClock")))
+        
+        self._eventLoop.call_soon_threadsafe(lambda: asyncio.create_task(self.callRing(alarmClock, timePoint)))
```
bernardnormier, line 30: This looks pretty complicated. There is not a simpler way to perform a "Task.Run"?

python/Ice/bidir/client/client.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) ZeroC, Inc.
+
+from EarlyRiser import AlarmClock, ButtonPressed
+
+class MockAlarmClock(AlarmClock):
```
bernardnormier, line 5: I would not call this file 'client.py'.

python/Ice/bidir/server/server.py
bernardnormier: and the file should not be server.py

python/Ice/bidir/client/client.py
```diff
@@ -0,0 +1,28 @@
...
+            self._stopPressed.set_result(None)
+
+    def ring(self, message, current):
+        print(f"Dispatching ring request {{ message = '{message}' }}")
+        if self._needMoreTime:
+            print(f"Returning {ButtonPressed.Snooze} to request more time.")
+            self._needMoreTime = False # we only snooze one time
+            return ButtonPressed.Snooze
+        else:
+            self._loop.call_soon_threadsafe(self.stop)
```
bernardnormier, line 26: As I understand it, we use asyncio for this communicator, so all dispatches are in the asyncio loop/thread.

The reason you call `self._loop.call_soon_threadsafe(self.stop)` and not self.stop is because you want to execute self.stop's continuation after the return? Why does it matter?

pepone: We don't know if a python dispatch method is sync or async until it returns. Dispatch cannot run in asyncio loop thread by default. If a dispatch method returns a coroutine, then we run it using the communicator's event loop adapter. I guess a way to avoid the `call_soon_threadsafe` call is to convert this dispatch method into an `async def`

python/Ice/bidir/README.md
pepone: yes

python/Ice/bidir/client/client.py
pepone: What about `clock.py`?

python/Ice/bidir/client/client.py
pepone: If we turn this dispatch into an async dispatch, we can use `call_soon` instead of `call_soon_threadsafe`. But either way we need to use this pattern, this is roughly equivalent to the `RunContinuationsAsynchronously` in the C# version. We don't want the stop pressed callback to run synchronously otherwise the dispatch would deadlock.

python/Ice/bidir/server/server.py
pepone: What about:

```
asyncio.run_coroutine_threadsafe(self.callRing(alarmClock, timePoint), self._eventLoop)
```

python/common/common/time.py
pepone: I guess something like:

```
(datetime.now() - datetime.combine(date.fromordinal(1), datetime.min.time())).total_seconds()
```

And then convert seconds to ticks.

python/Ice/bidir/server/server.py
pepone: Added comments, and rename it as `wakeUpService` not a great name because is also the base class name.

If we use the C# code as an example this should go into the `server` module.

Maybe something like this:

```
# in server/server/__init__.py

import BidirWakeUpService

# Move the servant definition to BidirWakeUpService.py

#in server.py

from server import BidirWakeUpService
```

But it might be overkill for a simple demo like this.

python/Ice/bidir/server/main.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+# Copyright (c) ZeroC, Inc.
+
+import Ice
+import asyncio
+from wakeUpService import BidirWakeUpService
```
bernardnormier, line 6: Maybe the convention should be like in the greeter demo: servant class goes into a file named after this class.

So Chatbot => chabot.py and later referenced as chatbot.Chatbot

BidirWakeUpService => bidir_wake_up_service.py and later referenced as bidir_wake_up_service.BidirWakeUpService

Or can you think of a better model?

pepone: or maybe just`greeter.py` `bidir.py` and import it as:

`from greeter import Chatbot`, `from bidir import BidirWakeUpService`

bernardnormier: I definitely prefer chatbot.py over greeter.py.

python/Ice/bidir/client/client.py
bernardnormier: I would prefer mock.py or mock_alarm_clock.py.

python/Ice/bidir/server/server.py
bernardnormier: Could be bidir.py.

python/Ice/bidir/server/server.py
bernardnormier: Better, but still pretty complicated.

python/common/common/time.py
bernardnormier: I would update the description. Something like:
```
Utility functions to convert between datetime and custom timestamps based on the Proleptic Gregorian calendar.
```

python/Ice/bidir/server/main.py
bernardnormier: What about bidir.py?

python/Ice/bidir/client/clock.py
```diff
@@ -0,0 +1,44 @@
...
+        self._eventLoop = eventLoop
+        self._stopPressed = eventLoop.create_future()
+
+    async def waitForStopPressed(self) -> None:
+        """
+        Asynchronously wait for the ring method to return ButtonPressed.Stop.
+        """
+        await self._stopPressed
+
+    async def ring(self, message: str, current: Ice.Current) -> ButtonPressed:
```
bernardnormier, line 22: I am confused here, and the reader may be confused too.

Does `ring` run in the event loop thread even if it's all synchronous?

python/Ice/bidir/slice/AlarmClock.ice
```diff
@@ -0,0 +1,27 @@
...
+        ButtonPressed ring(string message);
+    }
+
+    /// Represents the wake up service provided by the server.
+    interface WakeUpService
+    {
+        /// Schedules a call to the caller's {@link AlarmClock::ring} at the specified time.
+        /// @param timeStamp The time to ring the alarm clock. It's encoded as the number of ticks (100 nanoseconds)
+        /// since January 1, 0001 00:00:00 UTC in the Gregorian calendar.
+        void wakeMeUp(long timeStamp);
```
bernardnormier, line 25: Should we remap methods and parameters to snake_case?

python/Ice/bidir/client/clock.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    MockAlarmClock is an Ice servant that implements Slice interface AlarmClock
+    """
+
+    def __init__(self, eventLoop):
+        self._needMoreTime = True
+        self._eventLoop = eventLoop
+        self._stopPressed = eventLoop.create_future()
+
+    async def waitForStopPressed(self) -> None:
```
bernardnormier, line 16: should we use snake_case for our own methods?

python/Ice/bidir/client/clock.py
```diff
@@ -0,0 +1,44 @@
...
+from EarlyRiser import AlarmClock, ButtonPressed
+import Ice
+
+class MockAlarmClock(AlarmClock):
+    """
+    MockAlarmClock is an Ice servant that implements Slice interface AlarmClock
+    """
+
+    def __init__(self, eventLoop):
+        self._needMoreTime = True
```
bernardnormier, line 12: Maybe snake case would be more appropriate for Python fields?

bernardnormier (approved):

python/Ice/bidir/slice/AlarmClock.ice
pepone: not sure, what about class names? we also generate "Async" versions which would not use snake case.

python/Ice/bidir/server/server.py
pepone: I guess we can also make, `wakeMeUp` an `async def` method so that it runs in the event loop. And then use `_eventLoop.call_soon(self.callRing(alarmClock, timePoint)`.

PR #365, opened 2025-03-30
Add Python Context demo

pepone:


python/Ice/context/README.md
```diff
@@ -0,0 +1,74 @@
...
+```shell
+slice2py ../slice/Greeter.ice
+```
+
+- Run the server program:
+
+```shell
+python main.py
+```
+
```
bernardnormier, line 43: Should we add deactivate instructions? Not just here, but in all Python demos.

python/Ice/context/client/requirements.txt
```diff
@@ -0,0 +1,11 @@
+# Use ZeroC Nightly repository as the main index for pip
+--index-url https://download.zeroc.com/nexus/repository/pypi-nightly/simple/
+
+# Allow installing additional packages from PyPI main index if not found in the ZeroC repository
+--extra-index-url https://pypi.org/simple/
+
+# Allow installing pre-release versions (required for zeroc-ice nightly builds)
+--pre
+
+# Add the zeroc-ice package (latest nightly version)
```
bernardnormier, line 10: Maybe add something like:
```
# Add the zeroc-ice package (latest nightly version). This package includes the Slice compiler for Python (slice2py).
```

python/Ice/context/client/requirements.txt
```diff
@@ -0,0 +1,11 @@
+# Use ZeroC Nightly repository as the main index for pip
```
bernardnormier, line 1: I would at a period (`.`) at the end of all these comments.

python/Ice/context/client/main.py
```diff
@@ -0,0 +1,47 @@
...
+from VisitorCenter import GreeterPrx
+
+async def main():
+    # Set the Ice.ImplicitContext property to "Shared" before calling Ice.initialize.
+    # This is only necessary for the implicit context API (see below).
+    initData = Ice.InitializationData()
+    initData.properties = Ice.createProperties(sys.argv)
+    initData.properties.setProperty("Ice.ImplicitContext", "Shared")
+
+    # Configure the Ice runtime to use asyncio.
```
bernardnormier, line 19: While this is consistent with the comment on initialize below, I am concerned this is confusing.

The Ice runtime and an Ice communicator are not synonymous. For example, a "process logger" is definitely about the Ice runtime, while this asyncio enabling is only for a single commmunicator, not the Ice runtime as a whole.

For another PR!

bernardnormier (approved):

python/Ice/context/client/requirements.txt
pepone: Fine. I would use `the Slice to Python compiler` as that is the term we use more frequently.

PR #366, opened 2025-03-31
Add Python filesystem demo

pepone:


python/Ice/filesystem/client/main.py
```diff
@@ -0,0 +1,48 @@
...
+
+async def main():
+    # Create an Ice communicator to initialize the Ice runtime. It will be automatically destroyed at the end of the
+    # with block. We enable asyncio support by passing the current event loop to initialize.
+    with Ice.initialize(sys.argv, eventLoop=asyncio.get_running_loop()) as communicator:
+
+        # Create a proxy for the root directory.
+        rootDir = DirectoryPrx(communicator, "RootDir:tcp -h localhost -p 4061")
+
+        async def list_recursive(directory, depth=0):
```
bernardnormier, line 19: Can we move `list_recurvise` elsewhere?

I find it hard to follow to have this big local function in the middleware of the code.

python/Ice/filesystem/server/m_directory.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) ZeroC, Inc.
+
+from Filesystem import Directory, NodePrx
+import Ice
+
+class MDirectory(Directory):
+
```
bernardnormier, line 7: Missing doc-comment

python/Ice/filesystem/server/m_file.py
```diff
@@ -0,0 +1,42 @@
...
+        Parameters
+        ----------
+        name : str
+            The name of the file.
+        """
+        self._name = name
+        self._lines = []
+
+
+    def name(self, current: Ice.Current) -> str:
```
bernardnormier, line 24: A big part of this demo is to showcase implementation reuse with multiple inheritance - with a reusable MNode, like in C++. Can you do that in Python too? We can't do it in C# since we map Disp to classes and C# has no multiple inheritance for classes.

bernardnormier (approved):

python/Ice/filesystem/client/main.py
pepone: Moved the definition out before main.

python/Ice/filesystem/server/m_file.py
pepone: Reworked to use multiple inheritance.

bernardnormier (approved):
Looks good. The main README is still missing.

PR #367, opened 2025-03-31
Add Ice/config demo

pepone:


python/Ice/config/server/main.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+# Copyright (c) ZeroC, Inc.
+
+import Ice
+import chatbot
+import sys
+
+def main():
+    # Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the with
```
bernardnormier, line 9: In the client, we have a blank line after main and a blank like after the with statement. Which formatting is preferred?

bernardnormier (approved):
Looks good.

PR #369, opened 2025-03-31
Remove obsolete MATLAB demos

bernardnormier:
I kept only greeter (the new model) and matrix (unique to MATLAB ... but requires a Python server that was removed).

I also removed the Ice/gui demo (that unique to MATLAB): I don't think it's worthwhile to keep such GUI demos that don't show anything particular wrt Ice.


pepone (approved):

pepone:
> and matrix (unique to MATLAB ... but requires a Python server that was removed).

I find this demo isn't particularly good and showing any Ice features.

Ice for MATLAB doesn't have a special mapping for matrix types. Here we are showing how to create MATLAB matrix from a sequence.

The returned sequences randomly uses RowMajor or ColumnMajor, making the Python client suboptimal. It cannot take advantage of the builtin support for `numpy.ndarray`.

externl (approved):

PR #370, opened 2025-04-01
Add matrix demo

pepone:
This brings back the Ice/matrix python demo, just updated to use the new structure. I'm not convinced we should keep this demo.


bernardnormier:
If this a demo for MATLAB - that can be improved - I would actually put both the MATLAB client and Python server in the matlab demo tree. No Python client.

pepone:
> If this a demo for MATLAB 

I think we should remove the MATLAB side too, I don't see what Ice feature are we showing here or in the MATLAB side.

PR #372, opened 2025-04-01
Add Python cancellation demo

pepone:
Port the new C# cancellation demo to python.


python/Ice/cancellation/client/main.py
```diff
@@ -0,0 +1,61 @@
...
+            f = slowGreeter.greetAsync("bob")
+
+            async def cancel():
+                asyncio.sleep(4)
+                f.cancel()
+
+            asyncio.create_task(cancel())
+            greeting = await f
+            print(f"Received unexpected greeting: {greeting}")
+        except asyncio.exceptions.CancelledError as exception:
```
pepone, line 53: Note that because we are using asyncio futures we get `CancelledError` and not `InvocationCanceledException`

python/Ice/cancellation/client/main.py
```diff
@@ -0,0 +1,61 @@
...
+        slowGreeter = GreeterPrx(communicator, "slowGreeter:tcp -h localhost -p 4061")
+
+        # Create another slow greeter proxy with an invocation timeout of 4 seconds (the default invocation timeout is
+        # infinite).
+        slowGreeter4s = slowGreeter.ice_invocationTimeout(4000)
+
+        # Send a request to the regular greeter and get the response.
+        greeting = await greeter.greetAsync(getpass.getuser())
+        print(greeting)
+
```
bernardnormier, line 33: extra blank line

python/Ice/cancellation/client/main.py
```diff
@@ -0,0 +1,61 @@
...
+            greeting = await slowGreeter4s.greetAsync("alice")
+            print(f"Received unexpected greeting: {greeting}")
+        except Ice.InvocationTimeoutException as exception:
+            print(f"Caught InvocationTimeoutException, as expected: {exception}")
+
+        # Send a request to the slow greeter, and cancel this request after 4 seconds.
+        try:
+            f = slowGreeter.greetAsync("bob")
+
+            async def cancel():
```
bernardnormier, line 46: I would do simpler and just sleep and cancel inline. No need for a task.

python/Ice/cancellation/client/main.py
bernardnormier: I am not following.

If I await a future returned by Ice and the future is completed by some Ice local exception, I don't get that local exception??

python/Ice/cancellation/server/chatbot.py
```diff
@@ -0,0 +1,29 @@
...
+        ----------
+        delay : float
+            The delay before returning each greeting.
+        """
+        self._delay = delay
+
+    # Implements the method greet from the Greeter class generated by the Slice compiler.
+    # This variant is the synchronous implementation.
+    async def greet(self, name: str, current: Ice.Current) -> str:
+        print(f"Dispatching greet request {{ name = '{name}' }}")
```
bernardnormier, line 27: Dispatch cancellation is not doable in Python?

pepone: Doable yes. Not as elegant as with C# cancelation tokens. We can pass a future to the servant instead of a cancelation token.

Then instead of awaiting `asyncio.sleep` we would have to away both the sleep task and the extra future.

python/Ice/cancellation/client/main.py
pepone: > Does Ice for Python provide any wrapper for the cancellation function that Ice C++ returns? If yes, demo that. If not, don't demo canceling an asyncio future.

Here the cancelation is forwarded from `asyncio.Future -> Ice.Future -> C++ cancel callback`. The invocation is canceled when you cancel the asyncio.Future but you don't get the InvocationCanceled becase the future is already settled as canceled.

I would remove this part is not very interesting.

bernardnormier: This means the demo is now purely about invocation timeout? Then, we should rename it.

pepone: InvocationTimeout?

pepone:
renamed the demo to invocation_time, it no longer shows how to use future cancel API. And there is not dispatch cancellation.

python/Ice/cancellation/client/main.py
bernardnormier: Yes

python/Ice/invocation_timeout/README.md
```diff
@@ -0,0 +1,68 @@
+# Ice Invocation Timeout
+
+The Invocation Timeout demo shows how to use invocation timeouts
```
bernardnormier, line 3: missing period

python/Ice/invocation_timeout/server/chatbot.py
```diff
@@ -0,0 +1,29 @@
...
+
+        Parameters
+        ----------
+        delay : float
+            The delay before returning each greeting.
+        """
+        self._delay = delay
+
+    # Implements the method greet from the Greeter class generated by the Slice compiler.
+    # This variant is the synchronous implementation.
```
bernardnormier, line 25: This variant comment is not accurate.

bernardnormier (approved):
Looks good.

externl (approved):

PR #373, opened 2025-04-01
Add new context demo (MATLAB)

bernardnormier:


externl (approved):

matlab/Ice/context/client.m
```diff
@@ -0,0 +1,59 @@
...
+    % or IP address.
+    greeter = VisitorCenter.GreeterPrx(communicator, 'greeter:tcp -h localhost -p 4061');
+
+    % Create a request context.
+    context = containers.Map('KeyType', 'char', 'ValueType', 'char');
+
+    % Send a request to the remote object and get the response. We request a French greeting by setting 'language' in
+    % the context parameter.
+    context('language') = 'fr';
+    greeting = greeter.greet(char(java.lang.System.getProperty('user.name')), context);
```
pepone, line 44: I would hardcode the name if there ins't a plain MATLAB way to get this info, isn't that important.

matlab/Ice/context/client.m
```diff
@@ -0,0 +1,59 @@
+% Copyright (c) ZeroC, Inc.
+
+function client(args)
+    % We put the code generated by the Slice compiler in the 'generated' subdirectory.
+    addpath('generated');
```
pepone, line 5: Maybe rmpath would be better handled with a `onCleanup`, can we avoid the `addpath` by generating the code in the current directory, I think we still put everything in VisitorCenter, so is easy to .gitignore it.

matlab/Ice/context/client.m
```diff
@@ -0,0 +1,59 @@
...
+
+    % Load the Ice library if it is not already loaded.
+    if ~libisloaded('ice')
```
pepone, line 8: I guess we use this model because we don't call `unloadlibrary` we probably should.

matlab/Ice/context/client.m
```diff
@@ -0,0 +1,59 @@
...
+        loadlibrary('ice', @iceproto);
+    end
+
+    % The Slice module VisitorCenter maps to a MATLAB package with the same name.
+    import VisitorCenter.*
+
+    if nargin == 0
+        args = {};
```
pepone, line 16: Should the readme mention this args?

pepone (approved):
Looks good. Just a few minor things in the comments.

matlab/Ice/context/client.m
bernardnormier: I'd rather keep it for consistency with the demos in the other languages. It works fine and it's cross-platform.

matlab/Ice/context/client.m
bernardnormier: For a follow-up PR.

matlab/Ice/context/client.m
bernardnormier: I don't think we need to.

bernardnormier: Also, it doesn't work, so we should fix it!

PR #375, opened 2025-04-01
Update communicator creation comments in C# demos

bernardnormier:
There are essentially 3 variations:

a) Client
```
```

b) Server
```
// Create an Ice communicator. We'll use this communicator to create an object adapter.
```

c) Client+Server
```
// Create an Ice communicator. We'll use this communicator to create proxies, manage outgoing connections, and create
// an object adapter.
```

The plural for proxies and outgoing connections is intentional, even though many demos create a single proxy and a single outgoing connection. The communicator's job is to create proxies and manage outgoing connections (and more), and most client applications do create many proxies and sometimes multiples connections. 

On the other hand, I used singular for object adapter as it's very common for a server to create a single object adapter.


csharp/Ice/Config/Client/Program.cs
```diff
@@ -5,7 +5,7 @@
 // Slice module VisitorCenter in Greeter.ice maps to C# namespace VisitorCenter.
 using VisitorCenter;
 
-// Create an Ice communicator to initialize the Ice runtime. The communicator gets its configuration properties from
+// Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections. The communicator gets its configuration properties from
```
InsertCreativityHere, line 8: This line seems like it's way over the 120 limit.
I'd move `The communicator gets its...` onto the next line and consolidate from there.

csharp/IceStorm/Weather/Station/Program.cs
```diff
@@ -3,7 +3,8 @@
 using ClearSky;
 using System.Diagnostics;
 
-// Create an Ice communicator to initialize the Ice runtime. The communicator is disposed before the program exits.
+// Create an Ice communicator. We'll use this communicator to create proxies, manage  outgoing connections, and create
```
InsertCreativityHere, line 6: Extra space snuck in there.
```suggestion
// Create an Ice communicator. We'll use this communicator to create proxies, manage outgoing connections, and create
```

InsertCreativityHere (approved):
I like it!

pepone (approved):

externl (approved):

PR #376, opened 2025-04-01
Add new Ice/config demo for MATLAB

bernardnormier:
Ice parses the all the args but always treats the first one as the program name for logging, which results in odd logs:

```
-- 4/1/2025 16:34:45.975 --Ice.Trace.Network=2: Network: established ssl connection
   local address = 127.0.0.1:62513
   remote address = 127.0.0.1:4061
-- 4/1/2025 16:34:45.978 --Ice.Trace.Network=2: Network: closed ssl connection
   local address = 127.0.0.1:62513
   remote address = 127.0.0.1:4061
```

Here, `--Ice.Trace.Network=2` is treated as argv[0] == program name. Is this as expected?

### DIFF
--- a/matlab/Ice/context/client.m
+++ b/matlab/Ice/context/client.m
@@ -16,8 +16,7 @@ function client(args)
     % Set the Ice.ImplicitContext property to 'Shared' before calling Ice.initialize.
     % This is only necessary for the implicit context API (see below).
     initData = Ice.InitializationData();
-    [props, ~] = Ice.createProperties(args);
-    initData.properties_ = props;
+    initData.properties_ = Ice.createProperties(args);
     initData.properties_.setProperty('Ice.ImplicitContext', 'Shared');
 
     % Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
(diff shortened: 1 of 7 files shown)


matlab/Ice/context/client.m
```diff
@@ -16,11 +16,10 @@ function client(args)
     % Set the Ice.ImplicitContext property to 'Shared' before calling Ice.initialize.
     % This is only necessary for the implicit context API (see below).
     initData = Ice.InitializationData();
-    [props, ~] = Ice.createProperties(args);
-    initData.properties_ = props;
+    initData.properties_ = Ice.createProperties(args);
```
bernardnormier, line 19: Simplified this code. `createProperties(args)` and all initialize with args return 2 parameters, but we don't care about the second one.

pepone (approved):

pepone:
> Here, --Ice.Trace.Network=2 is treated as argv[0] == program name. Is this as expected?

The properties constructor assumes `argv[0]` is the program name, unless Ice.ProgramName is set. I guess we can set Ice.ProgramName in the config and add a comment about why this is necessary.

PR #381, opened 2025-04-02
Glacier2 callback demo

pepone:
This PR brings the new Glacier2 callback demo to Python.


python/Glacier2/callback/client/main.py
```diff
@@ -0,0 +1,64 @@
...
+from common.time import toTimestamp
+from datetime import datetime, timedelta
+
+async def main():
+    loop = asyncio.get_running_loop()
+    # Create an Ice communicator. We'll use this communicator to create proxies, manage outgoing connections, and
+    # create an object adapter. We enable asyncio support by passing the current event loop to initialize.
+    with Ice.initialize(sys.argv, eventLoop=loop) as communicator:
+
+        # Create a proxy to the Glacier2 router. The addressing information (transport, host and port number) is derived from
```
bernardnormier, line 19: Several comments go over the 120 column limit.

python/Glacier2/callback/client/main.py
```diff
@@ -0,0 +1,64 @@
...
+        # Obtain a category string from the router. We need to use this category for the identity of server->client callbacks
+        # invoked through the Glacier2 router.
+        clientCategory = await router.getCategoryForClientAsync();
+
+        # Create an object adapter with no name and no configuration, but with our router proxy. This object adapter is a
+        # "bidirectional" object adapter, like the one created by the Ice/Bidir client application. It does not listen on any
+        # port and it does not need to be activated.
+        adapter = communicator.createObjectAdapterWithRouter("", router);
+
+        # Register the MockAlarmClock servant with the adapter. The wake up service knows we use identity "alarmClock".
```
bernardnormier, line 40: The comment is incorrect here. We pass the proxy with the identity below.

python/Glacier2/callback/server/simple_wake_up_service.py
```diff
@@ -0,0 +1,61 @@
...
+        timestamp : int
+            The timestamp represented as the number of ticks since January 1, 0001 00:00:00 UTC.
+            Each tick is 100 nanoseconds.
+        current : Ice.Current
+            The Current object for the dispatch.
+        """
+        wakeUpDatetime = toDatetime(timestamp)
+        print(f"Dispatching wakeMeUp request {{ timeStamp = '{wakeUpDatetime.astimezone()}' }}")
+
+        # Schedule the callRing coroutine to run in the event loop. Ice dispatches async methods in the configured
```
bernardnormier, line 59: The second sentence sounds odd.

python/Glacier2/callback/client/main.py
```diff
@@ -0,0 +1,64 @@
...
+        mockAlarmClock = MockAlarmClock(loop)
+        alarmClock = AlarmClockPrx.uncheckedCast(
+            adapter.add(mockAlarmClock, Identity(name="alarmClock", category=clientCategory)))
+
+        # Start dispatching requests.
```
bernardnormier, line 45: Why do you activate this OA after writing "It does not listen on any port and it does not need to be activated."?

python/Glacier2/callback/client/main.py
```diff
@@ -0,0 +1,64 @@
...
+        # invoked through the Glacier2 router.
+        clientCategory = await router.getCategoryForClientAsync();
+
+        # Create an object adapter with no name and no configuration, but with our router proxy. This object adapter is a
+        # "bidirectional" object adapter, like the one created by the Ice/Bidir client application. It does not listen on any
+        # port and it does not need to be activated.
+        adapter = communicator.createObjectAdapterWithRouter("", router);
+
+        # Register the MockAlarmClock servant with the adapter. The wake up service knows we use identity "alarmClock".
+        mockAlarmClock = MockAlarmClock(loop)
```
bernardnormier, line 41: I am wondering if it would be cleaner to give MockAlarmClock a new future instead of the loop. Here, you get the impression MockAlarmClock keeps the loop.

python/Glacier2/callback/server/simple_wake_up_service.py
pepone: What about:

```
# Schedule the callRing coroutine in the current event loop. Ice dispatches async methods using the configured
# event loop, which can be accessed via asyncio.get_running_loop().
```

python/Glacier2/callback/client/main.py
pepone: Not sure. We should decide and fix bidir and callback to do the same.

All this `asyncio.get_running_loop()` are kind of an anti pattern and tie the servant to asyncio. Keeping the loop would be a better design, for example you can use another asyncio compatible loop (one that offers the same API).

python/Glacier2/callback/server/simple_wake_up_service.py
externl, line 59: I don't find it bad. 
```suggestion
        # Schedule the callRing coroutine to run in the asyncio event loop. Since Ice dispatches async methods in the configured
```

externl (approved):

python/Glacier2/callback/server/simple_wake_up_service.py
```diff
@@ -0,0 +1,61 @@
...
+        buttonPressed = await alarmClock.ringAsync("No more snoozing!")
+
+
+class SimpleWakeUpService(WakeUpService):
+    """
+    SimpleWakeUpService is an Ice servant that implements Slice interface WakeUpService.
+    """
+
+    async def wakeMeUp(self, alarmClock: AlarmClockPrx, timestamp: int, current: Ice.Current) -> None:
+        """
```
bernardnormier, line 43: In C# and other mappings, we don't add a doc-comment to the servant methods because we rely on "inheritdoc".
We just add a regular comment that doesn't repeat everything. Can we do that in Python too?

python/Glacier2/callback/config.glacier2
```diff
@@ -0,0 +1,18 @@
...
+
+# The server-visible endpoint of the Glacier2 router.
+# You need to configure this endpoint only when your servers make callbacks (call objects implemented by the clients).
+# If you don't use callbacks, don't set this property.
+Glacier2.Server.Endpoints=tcp -h 127.0.0.1
+
+# This Glacier router accepts any username/password combination.
+Glacier2.PermissionsVerifier=Glacier2/NullPermissionsVerifier
+
+# Turn-on tracing.
```
bernardnormier, line 14: Joe prefers "turn on"

python/Glacier2/callback/server/simple_wake_up_service.py
```diff
@@ -0,0 +1,61 @@
...
+        await asyncio.sleep(delay)
+
+    # First ring. This invocation reuses the connection established by the client.
+    buttonPressed = await alarmClock.ringAsync("It's time to wake up!")
+
+    # Keep ringing every 10 seconds until the user presses the stop button.
+    while buttonPressed is ButtonPressed.Snooze:
+        await asyncio.sleep(10)
+        buttonPressed = await alarmClock.ringAsync("No more snoozing!")
+
```
bernardnormier, line 35: I would print a message here like in the updated C# demo.

bernardnormier (approved):

python/Glacier2/callback/README.md
```diff
@@ -0,0 +1,82 @@
...
+```mermaid
+flowchart LR
+    c[Client<br>hosts AlarmClock] --bidir connection--> g[Glacier2 router:4063]
+    g --connection1--> s[Server:4061<br>hosts WakeUpService] --connection2--> g
+```
+
+We recommend running each program in a separate Python virtual environment. If you are new to Python virtual
+environments, see [Python Virtual Environments].
+
+## Running the server
```
bernardnormier, line 19: Running glacier2router is missing.

python/Glacier2/callback/server/simple_wake_up_service.py
pepone: Not sure, I will review this in a follow-up PR.

PR #382, opened 2025-04-02
Add new IceStorm/weather demo for MATLAB

bernardnormier:


csharp/IceStorm/Weather/Sensor/Program.cs
```diff
@@ -12,7 +12,7 @@
 // Create a proxy to the IceStorm topic manager.
 IceStorm.TopicManagerPrx topicManager = IceStorm.TopicManagerPrxHelper.createProxy(
     communicator,
-    "ClearSky/TopicManager:tcp -p 4061");
+    "ClearSky/TopicManager:tcp -p 4061 -h localhost");
```
bernardnormier, line 15: We should not showcase host-free direct proxies.

externl: We should update the other demos as well.

externl (approved):

pepone (approved):

PR #387, opened 2025-04-02
Add IceGrid demos for MATLAB

bernardnormier:
This PR adds 2 demos, greeter and locatorDiscovery.

### DIFF
--- a/csharp/IceGrid/LocatorDiscovery/Client/Program.cs
+++ b/csharp/IceGrid/LocatorDiscovery/Client/Program.cs
@@ -10,9 +10,7 @@ var initData = new Ice.InitializationData
 {
     properties = new Ice.Properties(ref args)
 };
-initData.properties.setProperty(
-    "Ice.Plugin.IceLocatorDiscovery",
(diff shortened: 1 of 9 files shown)


csharp/IceGrid/LocatorDiscovery/Client/Program.cs
```diff
@@ -10,9 +10,7 @@
 {
     properties = new Ice.Properties(ref args)
 };
-initData.properties.setProperty(
-    "Ice.Plugin.IceLocatorDiscovery",
-    "IceLocatorDiscovery:IceLocatorDiscovery.PluginFactory");
+initData.properties.setProperty("Ice.Plugin.LocatorDiscovery", "IceLocatorDiscovery:IceLocatorDiscovery.PluginFactory");
```
bernardnormier, line 13: We can simplify the plugin name.

externl (approved):

csharp/IceGrid/LocatorDiscovery/Client/Program.cs
bernardnormier: But we should not ... will fix in a follow-up PR.

PR #389, opened 2025-04-03
Add py-discovery greeter

pepone:
Port of the C# new discovery greeter demo to Python.


externl (approved):

python/IceDiscovery/greeter/client/main.py
```diff
@@ -0,0 +1,36 @@
...
+# Slice module VisitorCenter in Greeter.ice maps to Python module VisitorCenter.
+import VisitorCenter
+
+async def main():
+
+    # Configure the communicator to load the IceDiscovery plugin during initialization. This plugin installs a default
+    # locator on the communicator.
+    initData = Ice.InitializationData()
+    initData.properties = Ice.createProperties(sys.argv)
+    initData.properties.setProperty("Ice.Plugin.IceDiscovery", "IceDiscovery:createIceDiscovery")
```
bernardnormier, line 18: Does this work when Ice for Python is a single library that includes everything?

pepone: Yes, I tested with the pip package that has a single IcePy.so with everything.

Here for the static registered plugins all that matters is the plugin name.

> initData.properties.setProperty("Ice.Plugin.IceDiscovery", "1")

That would work as well. Maybe we should use this version as the entry point is not being used in this case.

This plug-in is always registered even when it is not bundled

https://github.com/zeroc-ice/ice/blob/80bed6974f84f871186e079f7c0d9d3a6335361d/python/modules/IcePy/Init.cpp#L136-L137

bernardnormier: I looked more closely and my understanding is:

- the name of the plugin generally does not matter, e.g. you could name is `Ice.Plugin.FooBar=IceDiscovery:createIceDiscovery` and it would work

- in particular, in C# and Java, the name of the plugin never matters

- the plugin properties names (as of Ice 3.8) are always fixed. Never prefixed on the plugin name.

- For special plugins in C++ registered via `Ice::registerXxx` (see Ice/RegistryPlugins.h), the name does matter and you better make sure you use the correct plugin name. This is especially important when loadOnInitialize = false.

For discovery and locator discovery, the plugin names are IceDiscovery and IceLocatorDiscovery. Since they matter (for this narrow C++ case), maybe we should use consistently IceDiscovery and IceLocatorDiscovery in all demos, in all languages?

PR #390, opened 2025-04-03
Switch to await using for communicator cleanup

bernardnormier:
This PR updates all the C# demos to use 'await using' for communicator cleanup.

Already before this PR, all Main were async:
 - client code always makes async invocations
 - server code always awaits shutdownCompleted

This PR also updates the MockAlarmClock to use a plain "run continuation synchronously" TCS. This means DisposeAsync actually starts destroying the communicator before dispatch (that returns Stop) completes. I verified this works fine by adding a temporary sleep before the return.

I also updated the WakeUpService to print a message when the "ring job" completes.


csharp/Ice/Bidir/Server/BidirWakeUpService.cs
```diff
@@ -45,6 +45,7 @@ public override void WakeMeUp(long timeStamp, Ice.Current current)
                 await Task.Delay(TimeSpan.FromSeconds(10));
                 buttonPressed = await alarmClock.RingAsync("No more snoozing!");
             }
+            Console.WriteLine("Client pressed Stop on alarm clock.");
```
bernardnormier, line 48: Makes it clearer the calls to RingAsync completed successfully.

csharp/IceStorm/Weather/Sensor/Program.cs
```diff
@@ -30,9 +30,10 @@
 // The proxy returned by createAsync and retrieveAsync is never null.
 Debug.Assert(topic is not null);
 
-// Create a Weather station proxy using the publisher proxy of the topic. The proxy returned by getPublisher is
-// never null.
-WeatherStationPrx weatherStation = WeatherStationPrxHelper.uncheckedCast(topic.getPublisher()!);
+// Create a WeatherStation proxy using the publisher proxy of the topic.
+Ice.ObjectPrx? publisher = await topic.getPublisherAsync();
```
bernardnormier, line 34: Switch to getPublisherAsync

pepone (approved):

externl (approved):

PR #393, opened 2025-04-03
Update Station code in IceStorm/Weather demo

bernardnormier:
Small updates to the csharp/IceStorm/Weather/Station code.


csharp/IceStorm/Weather/Station/Program.cs
```diff
@@ -11,9 +11,11 @@
 // listens on an OS-assigned TCP port, on all interfaces.
 Ice.ObjectAdapter adapter = communicator.createObjectAdapterWithEndpoints("StationAdapter", "tcp");
 
-// Register the ConsolePrinter servant with the adapter, and get a proxy to the new object. We use a UUID for the
-// identity because these subscribers (weather stations) are transient: when a subscriber shuts down, we want IceStorm
```
bernardnormier, line 15: This was confusing. IceStorm doesn't auto cleanup when a subscriber shuts down. It removes subscribers when it fails to deliver an event (based on the Retry QoS value), which is definitely not clear when you read the sentence above.

csharp/IceStorm/Weather/Station/Program.cs
```diff
@@ -38,14 +40,15 @@
...
 
-// Register our weather station with the topic.
-// subscribeAndGetPublisherAsync returns a publisher proxy that we don't need here.
-_ = await topic.subscribeAndGetPublisherAsync(theQoS: [], weatherStation);
-
 // Start dispatching requests.
 adapter.activate();
 Console.WriteLine("Listening...");
 
+// Register our weather station with the topic.
```
bernardnormier, line 47: Moved the call to subscribeAndGetPublisher _after_ activate. 

If activate takes a while (very unlikely here, most likely with IceGrid), we don't want IceStorm to fail to establish a connection to Station while activate is in progress. So best model is to subscribe after activate.

externl (approved):

csharp/IceStorm/Weather/Station/Program.cs
```diff
@@ -56,3 +59,7 @@
 
 // Wait until the communicator is shut down. Here, this occurs when the user presses Ctrl+C.
 await communicator.shutdownCompleted;
+
+// Remove the subscription to the topic.
+await topic.unsubscribeAsync(weatherStation);
```
pepone, line 64: We can mention the communicator shutdown doesn't affect the client-side functionality. I can look surprising to send a request after shutdown completed.

pepone (approved):

PR #394, opened 2025-04-03
Use "statically linked" plugins with MATLAB demos

bernardnormier:
This PR updates the IceDiscovery and IceLocatorDiscovery demos to use the (so-called) statically linked plugins.

### DIFF
--- a/matlab/IceDiscovery/greeter/client.m
+++ b/matlab/IceDiscovery/greeter/client.m
@@ -17,7 +17,7 @@ function client(args)
     % locator on the communicator.
     initData = Ice.InitializationData();
     initData.properties_ = Ice.createProperties(args);
-    initData.properties_.setProperty('Ice.Plugin.Discovery', 'IceDiscovery:createIceDiscovery');
+    initData.properties_.setProperty('Ice.Plugin.IceDiscovery', '1');
 
     % Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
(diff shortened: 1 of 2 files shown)


matlab/IceDiscovery/greeter/client.m
```diff
@@ -17,7 +17,7 @@ function client(args)
     % locator on the communicator.
     initData = Ice.InitializationData();
     initData.properties_ = Ice.createProperties(args);
-    initData.properties_.setProperty('Ice.Plugin.Discovery', 'IceDiscovery:createIceDiscovery');
```
pepone, line 20: Note this worked when using "Discovery" as the name because we ship the plug-in dlls next to the toolbox dll. Unlike with Python where we have a single IcePy.so.

pepone (approved):

pepone:
> (so-called) statically linked plugins.

Do we use this term? I think they are "registered' plugins, or linked but not need to be statically

bernardnormier:
> > (so-called) statically linked plugins.
> 
> Do we use this term? I think they are "registered' plugins, or linked but not need to be statically

See https://github.com/zeroc-ice/ice/blob/80bed6974f84f871186e079f7c0d9d3a6335361d/cpp/src/Ice/PluginManagerI.cpp#L350

The comment refers to "the static plugin factory table".  Indeed, we don't use the term "statically linked".

externl (approved):

PR #396, opened 2025-04-04
Add Python IceGrid greeter demo

pepone:


externl (approved):

python/IceGrid/greeter/README.md
```diff
@@ -0,0 +1,79 @@
+# IceGrid Greeter
+
+The IceGrid Greeter demo illustrates how to create a very simple IceGrid deployment that manages a Greeter server.
+
+Follow these steps to build and run the demo:
+
+1 Navigate to the server directory and create a Python virtual environment for running the server.
```
bernardnormier, line 7: Should be "1.".

We should also combine the server and icegridnode instructions. We should use the same virtual env for both.

bernardnormier (approved):

python/IceGrid/greeter/README.md
pepone: It is the same virtual env. Just activated twice.

PR #400, opened 2025-04-04
Add Java Context demo

pepone:
This PR adds Java Ice Context demo.


java/Ice/context/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -0,0 +1,43 @@
...
+        initData.properties.setProperty("Ice.ImplicitContext", "Shared");
+
+        // Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+        try (Communicator communicator = com.zeroc.Ice.Util.initialize(initData)) {
+            // Create a Greeter proxy. If you run the server on a different computer, replace localhost in the string
+            // below with the server's hostname or IP address.
+            GreeterPrx greeter = GreeterPrx.createProxy(communicator, "greeter:tcp -h localhost -p 4061");
+
+            // Send a request to the remote object and get the response. We request a French greeting by setting the
+            //  context parameter.
```
bernardnormier, line 28: extra space

java/Ice/context/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+> [!NOTE]
+> The Ice protocol does not support response contexts, only request contexts.
+
+## Building the Project
+
+This project uses Gradle and the [Application plugin]. To build the client and server applications, run:
+
+```shell
+./gradlew build
```
bernardnormier, line 16: I checked out your PR and this fails. greeter works fine.

```
./gradlew build
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
Caused by: java.lang.ClassNotFoundException: org.gradle.wrapper.GradleWrapperMain
```

java/Ice/context/client/src/main/java/com/zeroc/demos/greeter/Client.java
```diff
@@ -0,0 +1,43 @@
...
+            String greeting = greeter.greet(System.getProperty("user.name"), Map.of("language", "fr"));
+            System.out.println(greeting);
+
+            // Do it again, this time by setting the context on the proxy.
+            var greeterEs = GreeterPrx.uncheckedCast(greeter.ice_context(Map.of("language", "es")));
```
bernardnormier, line 33: Don't need the uncheckedCast in Java, C++ and more.

java/Ice/context/server/src/main/java/com/zeroc/demos/greeter/Chatbot.java
```diff
@@ -0,0 +1,29 @@
...
+import com.zeroc.demos.VisitorCenter.Greeter;
+import com.zeroc.Ice.Current;
+
+/**
+ * Chatbot is an Ice servant that implements Slice interface Greeter.
+ */
+class Chatbot implements Greeter
+{
+    // Implements the abstract method 'greet' from the Greeter interface generated by the Slice compiler.
+    // This variant is the synchronous implementation.
```
bernardnormier, line 14: Remove "This variant".

java/Ice/context/slice/GreeterAMD.ice
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) ZeroC, Inc.
```
bernardnormier, line 1: File not used

java/Ice/context/README.md
pepone: The gradle wrapper was missing. Added.

pepone:
A question for reviewers. What package should we use for client and server. Currently both use the same package, which doesn't seem like a best practice. They are still using `com.zeroc.demos.greeter`. 

I'm thinking:

`com.zeroc.demos.<demo-name>.<project-name>` for example `com.zeroc.demos.context.client` for the package name.

java/Ice/context/server/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+            // By default a Slice source set includes all Slice files in the srcDirs directories
+            // Here we override the default behavior by specifying the list of Slice files to include.
+            setIncludes(listOf("Greeter.ice"))
+        }
+    }
+}
+
+application {
+    // Specify the main entry point for the application.
+    mainClass.set("com.zeroc.demos.greeter.Server")
```
externl, line 22: ditto

java/Ice/context/server/src/main/java/com/zeroc/demos/greeter/Server.java
```diff
@@ -0,0 +1,45 @@
...
+    private static void shutdownCommunicatorOnCtrlC(Communicator communicator, Thread mainThread) {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            System.out.println("Caught Ctrl+C, shutting down...");
+            communicator.shutdown();
+
+            // Wait until the main thread completes.
+            try {
+                mainThread.join();
+            } catch (InterruptedException e) {
+                assert false : "Shutdown thread cannot be interrupted";
```
externl, line 41: Is this saying that it's impossible to get here as a shutdown thread cannot be interrupted, or that the user should not try to interrupted a shutdown thread?

externl (approved):

java/Ice/context/client/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+            // By default a Slice source set includes all Slice files in the srcDirs directories
+            // Here we override the default behavior by specifying the list of Slice files to include.
+            setIncludes(listOf("Greeter.ice"))
+        }
+    }
+}
+
+application {
+    // Specify the main entry point for the application.
+    mainClass.set("com.zeroc.demos.greeter.Client")
```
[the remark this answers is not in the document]
bernardnormier: Naturally, it should be named after the demo. Here the demo is Ice Context.

bernardnormier:
> `com.zeroc.demos.context.client` for the package name.

As Joe suggested, I would add the demo "dir" as well, so `com.zeroc.demos.ice.context.client`.

java/Ice/context/server/src/main/java/com/zeroc/demos/greeter/Server.java
bernardnormier: We create a new Thread in the addShutdownHook call. The application doesn't keep a reference to this thread and nowhere does it attempt to interrupt it.

Maybe we should have instead:
```java
assert false; // this can't happen
```

We have to catch InterruptedException because it's a checked exception.

java/Ice/context/server/src/main/java/com/zeroc/demos/greeter/Server.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.zeroc.demos.greeter;
+
+import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.ObjectAdapter;
+
+class Server {
+    public static void main(String[] args) {
+        // Create an Ice communicator. We'll use this communicator to create an object adapter.
+        try (Communicator communicator = com.zeroc.Ice.Util.initialize(args)) {
```
bernardnormier, line 11: Should we also import com.zeroc.Ice.Util?

java/Ice/context/slice/Greeter.ice
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+[["java:package:com.zeroc.demos"]]
```
bernardnormier, line 5: We should use the new `java:identifier` metadata, something like:

```
["java:identifier:com.zeroc.demos.visitor_center"]
```

Also,  file-metadata should never be "attached" to a module.

```

PR #403, opened 2025-04-04
Update IceGrid/secure `makedemocerts.py` to always set a password

### LINKED ISSUE #402: IceGrid/secure certificate issues
The 3.7 IceGrid/secure demo generates certificates without a password, however the config files specify a password.

externl:
We were previously relying the a behavior of `zeroc-icecertutils` that always set a password of `password`.  A recent updated change this behavior. The comment in the README was inaccurate, as we require a certificate. 

Fixes #402


pepone (approved):
Looks good. We should regenerate the top-level certs.

PR #405, opened 2025-04-07
JavaScript demos cleanup

pepone:
This PR is a small cleanup for JavaScript demos:

* Update the comments for communicator creation.
* Move timeUtil.ts to common package.
* Fix formatting with prettier.


js/common/package.json
externl, line 1: How does this work? Does the user need to `nom install` in here?

pepone: It doesn't. But without a package.json with type module. Node refuses to import this file as a module.

externl: What if you name the file `.mts`

bernardnormier (approved):

js/common/package.json
pepone: move timeUtil.ts back to each demo, sharing via a module is adding complexity here. I prefer to keep the build as simple as possible.

externl (approved):

PR #406, opened 2025-04-07
Add C# Ice/Secure demo

pepone:


csharp/Ice/Secure/Client/Program.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) ZeroC, Inc.
+
+// Slice module VisitorCenter in Greeter.ice maps to C# namespace VisitorCenter.
```
externl, line 3: I would move this comment next to the relevant using.  Maybe on the same line?

If this is what we do in the other demos then fine to leave as is.

csharp/Ice/Secure/Server/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    ServerCertificateContext = SslStreamCertificateContext.Create(
+        X509CertificateLoader.LoadPkcs12FromFile(
+            "../../../../certs/server.p12",
+            password: "password",
+            keyStorageFlags: X509KeyStorageFlags.Exportable),
+        additionalCertificates: null)
+};
+
+// Create an object adapter that listens for incoming requests and dispatches them to servants.
```
externl, line 20: I think we should mention the ssl transport.

"The object adapter endpoint uses the `ssl` transport ..."

externl (approved):
Looks good.  I think we should have comments on endpoints and TLS config options to say that TLS is used when using a "secure transport", or something to that affect.

csharp/Ice/Secure/Server/Program.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) ZeroC, Inc.
+
+// Create an Ice communicator. We'll use this communicator to create an object adapter.
```
bernardnormier, line 3: Needs to be next to the communicator creation.

csharp/Ice/Secure/Server/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+            "../../../../certs/server.p12",
+            password: "password",
+            keyStorageFlags: X509KeyStorageFlags.Exportable),
+        additionalCertificates: null)
+};
+
+// Create an object adapter that listens for incoming requests and dispatches them to servants.
+Ice.ObjectAdapter adapter = communicator.createObjectAdapterWithEndpoints(
+    "GreeterAdapter",
+    "ssl -p 4063",
```
bernardnormier, line 23: I would use 4061, just like in the Config demo. 4063/4 are for Glacier2. 4062 is for icerpc.

csharp/Ice/Secure/README.md
```diff
@@ -0,0 +1,27 @@
+# Ice Greeter
```
bernardnormier, line 1: Ice Secure

csharp/Ice/Secure/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```shell
+dotnet build
+```
+
+First start the Server program:
+
+```shell
+cd Server
+dotnet run -- --Ice.Trace.Network
```
bernardnormier, line 15: I am pretty sure just `dotnet run --Ice.Trace.Network` is sufficient.

csharp/Ice/Secure/Client/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+using VisitorCenter;
+
+// Create the authentication options with a custom certificate validation callback that uses the test certificates'
+// Root CA included with the demos.
+using var rootCA = new X509Certificate2("../../../../certs/cacert.der");
+var clientAuthenticationOptions = new SslClientAuthenticationOptions()
+{
+    RemoteCertificateValidationCallback = (sender, certificate, chain, errors) =>
+    {
+        if (certificate is X509Certificate2 peerCertificate)
```
bernardnormier, line 15: We need more comments here, this is too cryptic.

csharp/Ice/Secure/Client/Program.cs
pepone: Yes. The editor automatically added the other using directives and I didn't notice this was moved up.

csharp/Ice/Secure/Client/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+            // This ensures that only the root CA from the custom trust store is used for validation.
+            customChain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+            customChain.ChainPolicy.CustomTrustStore.Add(rootCA);
+
+            // Build returns true if the server certificate is valid according to the custom chain policy.
+            return customChain.Build(peerCertificate);
+        }
+        else
+        {
+            // Return false if the certificate is not of the expected type.
```
bernardnormier, line 36: Per https://learn.microsoft.com/en-us/dotnet/api/system.net.security.remotecertificatevalidationcallback?view=net-9.0

The certificate could be null apparently. Can it really have an unexpected type?

bernardnormier (approved):
Looks good.

csharp/Ice/Secure/Client/Program.cs
pepone: Updated the comment. I don't think it can be of a different type, .NET has this two certificate types but I think it is for backwards compatibility.

PR #415, opened 2025-04-10
Ruby demo addition: config

ayoitslilith:
This PR adds the `config` demo to Ruby's catalogue of demos. The code and explanations are directly based on the `greeter` demo for Ruby, but have been modified to utilize and explain the feature this is intended to showcase.


ruby/Ice/config/client.rb
```diff
@@ -0,0 +1,25 @@
...
+    # If you run the server on a different computer, replace localhost in the string below with the server's hostname
+    # or IP address.
+    properties = communicator.getProperties()
+    greeter = VisitorCenter::GreeterPrx.new(communicator, properties.getProperty("Greeter.Proxy"))
+    
+    # Send a request to the remote object and get the response.
+    greeting = greeter.greet(Etc.getlogin)
+
+    puts greeting
+end
```
externl, line 25: Missing newline at the end of this flie

ruby/Ice/config/client.rb
```diff
@@ -0,0 +1,25 @@
...
+# Copyright (c) ZeroC, Inc.
+
+require 'etc'
+require 'Ice'
+
+# Load the code generated by the Slice compiler. The Slice module VisitorCenter maps to a Ruby module with the same
+# name.
+require_relative 'Greeter.rb'
+
+# Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the block
```
bernardnormier, line 11: We changed the comment for communicator initialization. See for example the C# demos.

bernardnormier: The comment also need to say something about config.client. See for example the corresponding C# demo.

ruby/Ice/config/client.rb
```diff
@@ -0,0 +1,25 @@
...
+# returned by Ice::initialize.
+Ice::initialize(ARGV, "config.client") do |communicator|
+    # GreeterPrx is a class generated by the Slice compiler. We create a proxy from a communicator and a
+    # property found with the "Greeter.Proxy" specification with the address of the target object.
+    # If you run the server on a different computer, replace localhost in the string below with the server's hostname
+    # or IP address.
+    properties = communicator.getProperties()
```
bernardnormier, line 18: We should use `propertyToProxy` like in other language mappings.

ruby/Ice/config/client.rb
ayoitslilith: Funnily enough, I distinctly remember having the comments say something about config.client. I can't for the life of me remember what happened to that, but I'll have it fixed up in the next commit along with using `propertyToProxy()`.

ruby/Ice/config/client.rb
```diff
@@ -9,14 +9,15 @@
 require_relative 'Greeter.rb'
 
 # Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the block
-# returned by Ice::initialize.
+# returned by Ice::initialize. This communicator is initialized using command line input and the "config.client"
+# file as its arguments.
```
pepone, line 13: For the comment see:

 https://github.com/zeroc-ice/ice-demos/blob/39c3ecb81db7daa62b54c3298da542a3396bf7f1/csharp/Ice/Config/Client/Program.cs#L8-L11

ruby/Ice/config/client.rb
```diff
@@ -0,0 +1,26 @@
...
+require_relative 'Greeter.rb'
+
+# Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the block
+# returned by Ice::initialize. This communicator is initialized using command line input and the "config.client"
+# file as its arguments.
+Ice::initialize(ARGV, "config.client") do |communicator|
+    # GreeterPrx is a class generated by the Slice compiler. We create a proxy from a communicator, a property found
+    # with the "Greeter.Proxy" specification with the address of the target object, and the "config.client" file that
+    # we referenced during initialization.
+    # If you run the server on a different computer, replace localhost in the string below with the server's hostname
```
pepone, line 18: Need to update this comment, see C# version.

`replace localhost in the string...` that is from greeter but, here we use the config.

ruby/Ice/config/client.rb
```diff
@@ -0,0 +1,24 @@
...
+# name.
+require_relative 'Greeter.rb'
+
+# Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+# This communicator gets its configuration properties from file "config.client" in the client's current working directory.
+# The communicator initialization also parses the command-line options to find and set additional properties.
+Ice::initialize(ARGV, "config.client") do |communicator|
+    # GreeterPrx is a class generated by the Slice compiler. We create a proxy from a communicator, a property found
+    # with the "Greeter.Proxy" specification with the address of the target object, and the "config.client" file that
+    # we referenced during initialization.
```
pepone, line 17: I would use a simpler comment like in C#:

```suggestion
    # We create a Greeter proxy using the value of the Greeter.Proxy property in config.client.
```

pepone (approved):

ayoitslilith:
The CI is failing, but I don't believe this is because of any of my changes, considering that I did not modify anything in the C++ demos.

externl:
Yes, it's unrelated.

PR #416, opened 2025-04-10
Update IceDiscovery C# demos to use registerIceDiscovery

bernardnormier:
This PR updates the IceDiscovery demos for C# to use registerIceDiscovery like the C++ demos.

### DIFF
--- a/cpp/IceDiscovery/replication/Client.cpp
+++ b/cpp/IceDiscovery/replication/Client.cpp
@@ -12,11 +12,12 @@ using namespace std;
 int
 main(int argc, char* argv[])
 {
-    // Register the IceDiscovery plugin. The plugin will be loaded during communicator initialization and will
-    // install a default locator on the communicator.
+    // Register the IceDiscovery plugin. The plugin will be loaded during communicator initialization.
     Ice::registerIceDiscovery();
 
     // Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+    // The IceDiscovery plugin is created and initialized by initialize, and sets the default locator on the new
+    // communicator.
     Ice::CommunicatorPtr communicator = Ice::initialize(argc, argv);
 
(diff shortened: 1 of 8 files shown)


csharp/IceDiscovery/Greeter/Client/Program.cs
```diff
@@ -3,16 +3,13 @@
...
 
-// Configure the communicator to load the IceDiscovery plugin during initialization. This plugin installs a default
-// locator on the communicator.
-var initData = new Ice.InitializationData
-{
-    properties = new Ice.Properties(ref args)
-};
-initData.properties.setProperty("Ice.Plugin.Discovery", "IceDiscovery:IceDiscovery.PluginFactory");
+// Register the IceDiscovery plugin. The plugin will be loaded during communicator initialization.
+IceDiscovery.Util.registerIceDiscovery(loadOnInitialize: true);
```
bernardnormier, line 7: We should fix the C# code to make loadOnInitialize = true by default like in C++.

cpp/IceDiscovery/replication/Client.cpp
```diff
@@ -12,11 +12,12 @@ using namespace std;
...
 main(int argc, char* argv[])
 {
-    // Register the IceDiscovery plugin. The plugin will be loaded during communicator initialization and will
-    // install a default locator on the communicator.
+    // Register the IceDiscovery plugin. The plugin will be loaded during communicator initialization.
     Ice::registerIceDiscovery();
 
     // Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+    // The IceDiscovery plugin is created and initialized by initialize, and sets the default locator on the new
+    // communicator.
```
pepone, line 20: Seems a bit repetitive, the previous comment already says it will be loaded during communicator initialization.

Maybe just: `The IceDiscovery plugin sets the default locator on the new communicator.`

PR #419, opened 2025-04-15
Ruby demo addition: invocationTimeout

ayoitslilith:
This PR adds the InvocationTimeout demo to Ruby's catalogue of demos.


ruby/Ice/README.md
externl, line 1: This be list should be consolidated into a table like we've done in C++ and C# (for a future PR).

ruby/Ice/invocationTimeout/client.rb
externl, line 1: We should make sure this file (and others) have the executable bit set. Maybe we shouldn't set it and just remove the shebang. @pepone any feelings?

pepone: I prefer to keep seems standard practice. But if we do we should update instructions to use `./client.rb` rather than `ruby ./client.rb`

ruby/Ice/invocationTimeout/client.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'etc'
+require 'Ice'
+
+# Load the code generated by the Slice compiler. The Slice module VisitorCenter maps to a Ruby module with the same
+# name.
+require_relative 'Greeter.rb'
+
+# Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+# The communicator initialization parses the command-line options to find and set additional properties.
```
bernardnormier, line 12: The comments in this demo should be identical to the comments in the C#/MATLAB etc. demos, unless there is a compelling reason to have different comments.

ruby/Ice/invocationTimeout/client.rb
```diff
@@ -0,0 +1,35 @@
...
+Ice::initialize(ARGV) do |communicator|
+    # We create a Greeter proxy using a string that dictates how the proxy is meant to be set up.
```
bernardnormier, line 14: Same here: comments should be the same in all languages.

ruby/Ice/invocationTimeout/client.rb
```diff
@@ -0,0 +1,35 @@
...
+    # We create another proxy with an invocation timeout of four (4) seconds.
+    slowGreeter = VisitorCenter::GreeterPrx.new(communicator, "slowGreeter:tcp -h localhost -p 4061").ice_invocationTimeout(4000)
+
+    # We send a request and await a response with the regular greeter.
+    greeting = greeter.greet(Etc.getlogin)
+    puts greeting
+
+    # We also send a request with the slowGreeter and attempt to wait for a response.
+    begin
+        greeting = slowGreeter.greet(Etc.getlogin)
```
bernardnormier, line 26: Next call should be with 'alice'

ruby/Ice/invocationTimeout/client.rb
```diff
@@ -0,0 +1,35 @@
...
+        puts "Received unexpected greeting: " + greeting
+    rescue Ice::InvocationTimeoutException => exception
+        puts "Caught expected exception: " + exception.to_s
+    end
+
+    # We verify that the standard greeter still works.
+    greeting = greeter.greet("carol")
```
bernardnormier, line 33: bob after alice

ruby/Ice/invocationTimeout/client.rb
ayoitslilith: I was mainly keeping parity with the other Ruby comments, but I'll have it changed for the universal comments.

PR #420, opened 2025-04-15
Oops! Adds & modifies README.md files for the Ruby Config demo

ayoitslilith:
Looks like I forgot to add `ruby/Ice/config/README.md` and adjust `./ruby/Ice/README.md` for the Config demo. This PR is intended to rectify this, included with a simple explanation of the purpose of the Config demo.

### DIFF
--- a/ruby/Ice/config/README.md
+++ b/ruby/Ice/config/README.md
@@ -1,0 +1,20 @@
+# Config
+
+The Config demo shows off how to use a config file to create a client proxy.
+
+Ice for Ruby supports only client-side applications. As a result, you first need to start a Config server implemented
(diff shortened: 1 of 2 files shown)


ruby/Ice/config/README.md
```diff
@@ -0,0 +1,20 @@
+# Config
+
+The Config demo shows off how to use a config file to create a client proxy.
```
externl, line 3: We should use the wording from the MATLAB demo. 

```suggestion
This demo shows how to configure a client application using an Ice configuration file.
```

pepone (approved):

externl (approved):

PR #426, opened 2025-04-21
Ruby demo addition: IceDiscovery greeter

ayoitslilith:
This PR adds the IceDiscovery greeter demo to Ruby's catalogue of demos.

Because of some quirks concerning IceDiscovery, this Ruby client was tested with a C# server, both run in WSL.


ruby/IceDiscovery/greeter/README.md
```diff
@@ -0,0 +1,20 @@
+# IceDiscovery Greeter
+
+This demo illustrates how to use the IceDiscovery plugin with Ice for Ruby.
+
+Ice for Ruby supports only client-side applications. As a result, you first need to start a Greeter server implemented
```
pepone, line 5: Should this say "an IceDiscovery Greeter server ..." a plain Greeter would not work with this client.

ruby/IceDiscovery/greeter/client.rb
```diff
@@ -0,0 +1,29 @@
...
+# name.
+require_relative 'Greeter.rb'
+
+# Configure the communicator to load the IceDiscovery plugin during initialization. This plugin installs a default
+# locator on the communicator.
+initData = Ice::InitializationData.new()
+initData.properties = Ice.createProperties(ARGV)
+initData.properties.setProperty("Ice.Plugin.IceDiscovery", "1")
+
+# Create an Ice communicator to initialize the Ice runtime. The communicator is destroyed at the end of the block
```
pepone, line 17: The communicator comment needs to be updated.

ayoitslilith: I have now changed this to read the same as the Python and MATLAB equivalents.

pepone (approved):

PR #430, opened 2025-04-24
Ruby demo addition: IceStorm weather

ayoitslilith:
This PR adds the weather demo for IceStorm to Ruby's catalogue of demos. This relies on the latest version of the Ice Ruby gem to function, and was tested using a locally built version of the gem.


ruby/IceStorm/weather/WeatherStation.rb
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) ZeroC, Inc.
```
pepone, line 1: Generated code. Remove and add it to .gitignore

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env ruby
+# Copyright (c) ZeroC, Inc.
+
+require 'date'
+require 'etc'
+require 'Ice'
+require 'IceStorm'
+
+# The Slice module ClearSky in WeatherStation.ice maps a Ruby module with the same name.
```
pepone, line 9: This comment is a bit odd for Ruby, in other mappings we have a similar comment for `using` or module imports. But here `ClearSky` doesn't appear anywhere.

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
...
+    # and then retrieve the proxy if the topic was already created.
+    topicName = 'weather'
+    begin
+        topic = topicManager.create(topicName)
+    rescue IceStorm::TopicExists => ex
+        topic = topicManager.retrieve(topicName)
+    end
+
+    # Create a WeatherStation proxy using the publisher proxy of the topic. The proxy returned by getPublisher is
+    # never null.
```
pepone, line 28: is `nil` for Ruby, ins't it?

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
...
+    # Generate a sensor ID for this weather sensor.
+    sensorId = "sensor-" + rand(1000).to_s
+
+    # Send a reading every second to the weather station(s) via IceStorm. We keep sending until the user presses Ctrl+C.
+    puts sensorId + " reporting. Press Ctrl+C to stop..."
+
+    while true
+        # Generate a random temperature and humidity reading.
+        temperature = (190 + rand(40)) / 10; # Temperature in degrees Celsius (19.0 to 23.0).
+        humidity = (450 + rand(100)) / 10; # Humidity in percentage (45.0 to 55.0).
```
pepone, line 40: We should use Float numbers. `190.0` ...

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
...
+        # Send the reading to the weather station(s) via IceStorm.
+        timeStamp = DateTime.now
+        weatherStation.report(sensorId, timeStamp.to_s, reading)
+
+        # Wait for one second before sending the next reading.
+        sleep(1)
+    end
+
+    
+end
```
pepone, line 54: missing new line

ruby/IceStorm/weather/sensor.rb
ayoitslilith: Clearsky appears at lines 29 and 43, and is taken from WeatherStation.rb, which was generated off of WeatherStation.ice.  The comment is somewhat oddly worded, but it has been this way consistently in the Ruby demos as to maintain parity with other Ruby demo equivalents and other language equivalents.

ruby/IceStorm/weather/sensor.rb
ayoitslilith: You are correct, I'll change this shortly.

ruby/IceStorm/weather/sensor.rb
pepone: I know this is not new, I mean is odd in general for Ruby demos to have this comment.

For example in C# we have:

https://github.com/zeroc-ice/ice-demos/blob/21fbe7e3c43a47fa280d0834665fa4aedae1ddc4/csharp/Ice/Greeter/Client/Program.cs#L3-L4

At the point we add the using, and it is clear what `VisitorCenter` is.

In JavaScript:

https://github.com/zeroc-ice/ice-demos/blob/21fbe7e3c43a47fa280d0834665fa4aedae1ddc4/js/Ice/context/client.ts#L4

We don't but It would make sense to have one there.

For Python is similar to C#

https://github.com/zeroc-ice/ice-demos/blob/21fbe7e3c43a47fa280d0834665fa4aedae1ddc4/python/Ice/greeter/client/main.py#L9-L10

Anyway we can review this in a follow up PR, for me the comment is somehow unrelated to the required statement where it is placed.

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
...
+        reading = ClearSky::AtmosphericConditions.new(temperature, humidity)
+
+        # Send the reading to the weather station(s) via IceStorm.
+        timeStamp = DateTime.now
+        weatherStation.report(sensorId, timeStamp.to_s, reading)
+
+        # Wait for one second before sending the next reading.
+        sleep(1)
+    end
+
```
pepone, line 52: extra new line here?

pepone (approved):

ruby/IceStorm/weather/sensor.rb
```diff
@@ -0,0 +1,54 @@
...
+    # Create a WeatherStation proxy using the publisher proxy of the topic. The proxy returned by getPublisher is
+    # never nil.
+    weatherStation = ClearSky::WeatherStationPrx::uncheckedCast(topic.getPublisher())
+
+    # Generate a sensor ID for this weather sensor.
+    sensorId = "sensor-" + rand(1000).to_s
+
+    # Send a reading every second to the weather station(s) via IceStorm. We keep sending until the user presses Ctrl+C.
+    puts sensorId + " reporting. Press Ctrl+C to stop..."
```
externl, line 35: We should see if ruby has good signal handling we can use.

externl (approved):

PR #433, opened 2025-04-24
CMakeLists.txt updates

externl:
- Use PRIVATE scope linking
- Copy ICE_RUNTIME_DLLS in POST_BUILD command

### DIFF
--- a/cpp/DataStorm/clock/CMakeLists.txt
+++ b/cpp/DataStorm/clock/CMakeLists.txt
@@ -5,7 +5,7 @@ project(datastorm_clock CXX)
 include(../../cmake/common.cmake)
 
 add_executable(reader Reader.cpp)
-target_link_libraries(reader PRIVATE Ice::Ice Ice::DataStorm)
+target_link_libraries(reader Ice::Ice Ice::DataStorm)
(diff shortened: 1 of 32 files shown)


cpp/DataStorm/clock/CMakeLists.txt
```diff
@@ -5,15 +5,19 @@ project(datastorm_clock CXX)
 include(../../cmake/common.cmake)
 
 add_executable(reader Reader.cpp)
-target_link_libraries(reader Ice::Ice Ice::DataStorm)
+target_link_libraries(reader PRIVATE Ice::Ice Ice::DataStorm)
```
pepone, line 8: What does PRIVATE means here?

pepone (approved):

cpp/DataStorm/clock/CMakeLists.txt
externl: It means link with the library but don't include it int he libraries link list. Like bzip2 in Ice. For executables it doesn't really matter you put there as nothing links with an executable. It's more signaling intent. 

For a library it's really about if the library exposes APIs from its dependencies. https://leimao.github.io/blog/CMake-Public-Private-Interface/

Issue #425, opened 2025-04-20
Add user exception demo


bernardnormier:
We should demonstrate user-defined Slice exceptions in a new demo.

See https://github.com/icerpc/icerpc-csharp/tree/main/examples/slice/CustomError

PR #436, opened 2025-04-29
C# demo addition: CustomError

ayoitslilith:
This PR adds the C# CustomError demo from IceRPC to Ice's catalogue. The general structuring of this demo is primarily based on the C# Ice Greeter demo, though I referenced some other demos that used Exceptions.


ayoitslilith:
Forgot to mention - this is meant to address #425

csharp/Ice/CustomError/Client/Program.cs
```diff
@@ -0,0 +1,30 @@
...
+
+foreach (string name in names)
+{
+    // Send a request to the remote object and get the response.
+    try
+    {
+        greeting = await greeter.GreetAsync(name);
+        Console.WriteLine(greeting);
+    }
+    catch (GreeterException grExcept)
```
bernardnormier, line 26: We don't use complete words like greeterException or exception.

csharp/Ice/CustomError/CustomError.sln
```diff
@@ -0,0 +1,48 @@
...
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Client", "Client\Client.csproj", "{4AA07808-8EBD-49BA-A2B3-CF69DE7B9062}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Server", "Server\Server.csproj", "{C885F785-52A0-42A3-8C38-21940252BE5E}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
bernardnormier, line 12: We should have only Any CPU. Simplest is to copy and modify an existing solution file.

csharp/Ice/CustomError/Server/Chatbot.cs
```diff
@@ -0,0 +1,34 @@
...
+        if (name.Length > MaxLength)
+        {
+            throw new GreeterException("Name is longer than maximum!", GreeterError.NameTooLong);
+        }
+        switch (name)
+        {
+            case "":
+                throw new GreeterException("Name is empty!", GreeterError.EmptyName);
+            case "jimmy":
+                Thread.Sleep(5000);
```
bernardnormier, line 28: This does not make sense: we block for 5s and then throw the exception.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+module VisitorCenter
+{
+    /// The errors returned by the Greeter.
+    enum GreeterError { EmptyName, NameTooLong, Away };
+
+    /// Represents an exception.
+    exception GreeterException {
+        /// The reason for the exception.
```
bernardnormier, line 12: You should call this field message.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,28 @@
...
+        string reason;
+
+        //The ID for the exception, given as a GreeterError enum.
```
bernardnormier, line 15: Always `///` followed by a space.

And this error is not an ID, but an error code.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+module VisitorCenter
+{
+    /// The errors returned by the Greeter.
+    enum GreeterError { EmptyName, NameTooLong, Away };
+
+    /// Represents an exception.
```
bernardnormier, line 10: I would improve this comment - this is the central part of this demo.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+module VisitorCenter
+{
+    /// The errors returned by the Greeter.
```
bernardnormier, line 7: The greeter does not return these enumerators. This enum is a field of GreeterException.

csharp/Ice/CustomError/Server/Chatbot.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace CustomErrorServer;
+
+/// <summary>Chatbot is an Ice servant that implements Slice interface Greeter.</summary>
+internal class Chatbot : GreeterDisp_
+{
+    private const int MaxLength = 7;
+
+    /// <inheritdoc/>
+    // Implements the abstract method Greet from the GreeterDisp_ class generated by the Slice compiler.
+    // This variant is the synchronous implementation.
```
bernardnormier, line 14: We don't provide an async variant, so I would remove this comment.

csharp/Ice/CustomError/Server/Chatbot.cs
```diff
@@ -0,0 +1,33 @@
...
+    public override string Greet(string name, Ice.Current current)
+    {
+        Console.WriteLine($"Dispatching greet request {{ name = '{name}' }}");
+
+        if (name.Length > MaxLength)
```
bernardnormier, line 19: You should use a switch expression, just like:
https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/examples/slice/CustomError/Server/Chatbot.cs#L23

```
return name switch
{
   ...
}
```

and you can throw after `=>`.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,31 @@
...
+        /// The ID for the exception, given as a GreeterError enum.
+        GreeterError error;
+    }
+
+    /// Represents a simple greeter.
+    interface Greeter
+    {
+        /// Creates a personalized greeting.
+        /// @param name The name of the person to greet.
+        /// @return The greeting or exception.
```
pepone, line 27: We should add a `@throws` comment for the exception.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+module VisitorCenter
+{
+    /// This enum is used as a field of GreeterException, functionally as an error ID.
```
pepone, line 7: I think "error code" is a more common term.

csharp/Ice/CustomError/Server/Chatbot.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    private const int MaxLength = 7;
+
+    /// <inheritdoc/>
+    // Implements the abstract method Greet from the GreeterDisp_ class generated by the Slice compiler.
+    public override string Greet(string name, Ice.Current current)
+    {
+        Console.WriteLine($"Dispatching greet request {{ name = '{name}' }}");
+
+        GreeterException? exception = name switch
```
bernardnormier, line 18: What I meant was:
```
return name switch
{
   "" => throw new GreeterException("Name is empty!", GreeterError.EmptyName),
   ...
    _ => $"Hello, {name}!";
};
```

bernardnormier (approved):
I'll improve the Slice comments and README once this PR is merged.

pepone (approved):

csharp/Ice/CustomError/README.md
```diff
@@ -0,0 +1,27 @@
+# Custom error
+
+The Custom error example shows how to define an operation that throws an exception defined as
+`GreeterException` when an issue is present.
+
+In this example, `GreeterException` holds `reason` and `error`, values defined as a string and a
+`GreeterError` meant to hold an error message and an error ID.
```
InsertCreativityHere, line 7: Personally, I would remove this sentence.
Not because it's wrong, but because it's discussing the internals of `GreeterException`, which (to me) isn't a critical part of the demo. What's important is that you define an exception in Slice, and can throw it.
What information that exception happens to hold is secondary, and will vary.

To be clear, I'm **not** saying that this shouldn't be documented _anywhere_.
It's documented inside the Slice file itself, and that is a very good thing. I just don't think it's critical enough that we should explain it a _second time_ in the main README here. Ya feel?

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,32 @@
...
+
+module VisitorCenter
+{
+    /// This enum is used as a field of GreeterException, functionally as an error code.
+    enum GreeterError { EmptyName, NameTooLong, Away };
+
+    /// Represents an exception.
+    /// It contains a `string` message and `GreeterError` error.
+    /// The message is used by any application attempting to throw an exception to describe the nature of the exception.
+    /// The error is used similarly, but is limited to the enums in GreeterError and is better for internal identifying.
```
InsertCreativityHere, line 13: I see that this was all added in response to Bernard's review comment.
But I don't think this is the "right way" to of 'improved' it.

There is one critical piece of information which should always be mentioned in an exception:
"what kind of failure does this represent"/"what causes this exception to be thrown".

So, I wouldn't bother mentioning the what fields this exception has. A user can see which fields it contains by looking down a couple lines. And the parts explaining what the `message` and `error` are used for also shouldn't be here IMO. They should be on the comments for the fields themselves.
This is the comment for the _exception_ itself, not for it's _fields_.

TL;DR, I'd remove these lines about the fields (maybe moving the explanations to the fields themselves),
and in their place, describe what this exception is about or what causes it to be thrown, ie. invalid names.

csharp/Ice/CustomError/slice/Greeter.ice
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+module VisitorCenter
+{
+    /// This enum is used as a field of GreeterException, functionally as an error code.
```
InsertCreativityHere, line 7: I think something like this would be better:
```suggestion
    /// This enum is used to specify the underlying cause of a GreeterException.
```
It conveys the same information, but more succinctly. And mentioning that it's a field is an unnecessary level of specificity, and doesn't say anything knew. You can already know this by just reading the source code. So, no reason to quote-unquote "restate the obvious".

Again, feel free to tweak my wording if you really preferred "error code" over "cause". I just think that saying "cause" is simpler and has less "implied baggage" compared to an error code (because usually error codes are integers, not enums).

InsertCreativityHere (commented):
All the code looks fine!
But I had some thoughts on the comments/documentation.

I tried to write out my thoughts for clarity's sake, and so you understand the reasoning behind it. Sorry if they're a little too wordy.

csharp/Ice/CustomError/README.md
```diff
@@ -0,0 +1,27 @@
+# Custom error
+
+The Custom error example shows how to define an operation that throws an exception defined as
+`GreeterException` when an issue is present.
```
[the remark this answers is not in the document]
ayoitslilith: The only other use of the phrase "examples" (that I'm aware of) is present in IceRPC - particularly, the CustomError demo this comes from. A lot of the weird steps away from conventions here are due to me following conventions established in that demo - which, as much as I tried to move away from them, it was difficult to remain fully divorced from the conventions of the demo I was referencing.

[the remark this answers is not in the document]
InsertCreativityHere: > as much as I tried to move away from them, it was difficult to remain fully divorced from the conventions of the demo I was referencing.

Understandably so!

InsertCreativityHere (approved):
Looks good to me.

Bernard had one other suggestion still, about the exact wording we conventionally use for `@throws`. You should fix it, but this is just what we've decided to use for Slice, not a generally true piece of advice. https://github.com/zeroc-ice/ice-demos/pull/436#discussion_r2069322233

PR #439, opened 2025-04-29
Update MATLAB demos to use lowercase namespace + no import

bernardnormier:
No MATLAB demo shows unmarshaling class or exception instance, and as a result, no demo needs a SliceLoader to deal with the new namespace.

### DIFF
--- a/matlab/Glacier2/greeter/Greeter.ice
+++ b/matlab/Glacier2/greeter/Greeter.ice
@@ -2,6 +2,7 @@
 
 #pragma once
 
+["matlab:identifier:visitorcenter"]
 module VisitorCenter
 {
     /// Represents a simple greeter.
--- a/matlab/Glacier2/greeter/client.m
+++ b/matlab/Glacier2/greeter/client.m
@@ -1,9 +1,6 @@
 % Copyright (c) ZeroC, Inc.
 
 function client(args)
-    % The Slice module VisitorCenter maps to a MATLAB namespace with the same name.
-    import VisitorCenter.*
-
     if nargin == 0
         args = {};
     end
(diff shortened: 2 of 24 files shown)


matlab/Glacier2/greeter/Greeter.ice
```diff
@@ -2,6 +2,7 @@
 
 #pragma once
 
+["matlab:identifier:visitorcenter"]
```
bernardnormier, line 5: namespaces are usually lowercase in MATLAB. It's unfortunate we can't easily rename namespace Ice, IceStorm etc.

matlab/Glacier2/greeter/client.m
```diff
@@ -1,9 +1,6 @@
 % Copyright (c) ZeroC, Inc.
 
 function client(args)
-    % The Slice module VisitorCenter maps to a MATLAB namespace with the same name.
-    import VisitorCenter.*
```
bernardnormier, line 5: import namespace.* is discouraged in MATLAB. Here, it's clearer and simpler to not rely on import.

matlab/Ice/context/client.m
```diff
@@ -1,11 +1,8 @@
 % Copyright (c) ZeroC, Inc.
 
 function client(args)
-   % The Slice module VisitorCenter maps to a MATLAB namespace with the same name.
-   import VisitorCenter.*
-
    if nargin == 0
-       args = {};
+        args = {};
    end
```
pepone, line 6: The `if` `end` indentation is missing one space.

pepone (approved):

matlab/Ice/greeter/Greeter.ice
```diff
@@ -2,6 +2,9 @@
 
 #pragma once
 
+// We use matlab:identifier to get a lowercase namespace in MATLAB; without this metadata directive, the mapped
+// namespace is VisitorCenter.
+["matlab:identifier:visitorcenter"]
```
InsertCreativityHere, line 7: This is a very minor suggestion.
```suggestion
// We use matlab:identifier to get a lowercase namespace in MATLAB; without this metadata directive, the mapped
// namespace would be VisitorCenter.
["matlab:identifier:visitorcenter"]
```

InsertCreativityHere (approved):
Looks good to me!

PR #441, opened 2025-04-30
Add IceGrid/greeter mermaid diagram

externl:
This is my proposed diagram. I initially had a much more complex one but slimed it down to this.

```mermaid
flowchart LR
    Client(Client) --> | locate greeter | Locator
    subgraph Registry[IceGrid Registry]
        direction LR
        Locator[locator:4061]
    end
    Node[IceGrid Node]
    Server(Server<br/>hosts Chatbot)
    Registry <--> Node --> | activate | Server
    Locator -.-> |return greeter endpoints | Client
    Client ==> |greet request| Server
```

Once we settle on something  I'll copy to all demos.

### DIFF
--- a/cpp/IceGrid/greeter/README.md
+++ b/cpp/IceGrid/greeter/README.md
@@ -5,19 +5,14 @@ The IceGrid Greeter demo illustrates how to create a very simple IceGrid deploym
 ```mermaid
 flowchart LR
     Client(Client) --> | locate greeter | Locator
-
     subgraph Registry[IceGrid Registry]
         direction LR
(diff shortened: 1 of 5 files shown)


pepone:
Looks good, regarding terminology, what about using "activate" instead of start, and "return greeter endpoints" instead of "return greeter proxy".

bernardnormier:
In the server, I would write 'hosts Chatbot'. Otherwise, looks good.

bernardnormier (approved):

pepone (approved):

PR #443, opened 2025-05-01
Add cpp/IceStorm/weather

externl:


cpp/IceStorm/weather/ConsolePrinter.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) ZeroC, Inc.
+
+#ifndef CONSOLE_PRINTER1_H
+#define CONSOLE_PRINTER1_H
```
pepone, line 4: Why `1_H`?

cpp/IceStorm/weather/Sensor.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <random>
+#include <thread>
+
+using namespace ClearSky;
+using namespace std;
+
+int
+main(int argc, char* argv[])
+{
+    // CtrlCHandler must be called before the node is created or any other threads are started.
```
pepone, line 19:
```suggestion
    // CtrlCHandler must be created before the communicator is created or any other threads are started.
```?

File: cpp/IceStorm/weather/Sensor.cpp [hunk 3]
```diff
@@ -0,0 +1,98 @@
...
+
+    // Make sure the communicator is destroyed at the end of this scope.
+    Ice::CommunicatorHolder communicatorHolder{communicator};
+
+    // Create a proxy to the IceStorm topic manager.
+    IceStorm::TopicManagerPrx topicManager{communicator, "ClearSky/TopicManager:tcp -p 4061 -h localhost"};
+
+    // Retrieve a proxy to the "weather" topic: we first create a topic with the given name (in case we are the first),
+    // and then retrieve the proxy if the topic was already created.
+    std::optional<IceStorm::TopicPrx> topic;
```
pepone, line 33: Use `const std::string topicName = "weather";` like in the Station.

pepone (approved):

cpp/IceStorm/weather/ConsolePrinter.h
externl: From a bad copy and paste

cpp/IceStorm/weather/Sensor.cpp
```diff
@@ -0,0 +1,99 @@
...
+    ctrlCHandler.setCallback(
+        [&running](int)
+        {
+            running = false;
+            std::cout << "Shutting down..." << std::endl;
+        });
+
+    // Initialize random number generators.
+    std::random_device rd;
+    std::mt19937 gen{rd()};
```
bernardnormier, line 75: Merge with previous line

cpp/IceStorm/weather/Sensor.cpp
```diff
@@ -0,0 +1,99 @@
...
+    assert(weatherStation->ice_isTwoway());
+
+    // Generate a sensor ID for this weather station.
+    string uuid = Ice::generateUUID();
+    string sensorId = "sensor-" + uuid.substr(uuid.length() - 4);
+
+    cout << sensorId << " reporting. Press Ctrl+C to stop." << endl;
+
+    // Shutdown the sensor on Ctrl-C.
+    std::atomic<bool> running{true};
```
bernardnormier, line 65: I think it would be more appropriate to use a promise and no sleep_for.

Ctrl+C sets the result and clears itself (to avoid double set). The loop waits for the future to become available (for 1 second).

cpp/IceStorm/weather/WeatherStation.ice
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+module ClearSky
+{
+    /// Represents the atmospheric conditions measured by a sensor.
+    class AtmosphericConditions
```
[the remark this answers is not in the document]
externl: We decided for the previous demo that uses WeatherStation to use the built-in printing that we've added for 3.8.

cpp/IceStorm/weather/Sensor.cpp
externl: That's so much more complicated and unnecessary IMO. There's no treason to use promises for everything.

bernardnormier: It's not really more complicated and in all other demos the "sleep" gets interrupted by Ctrl+C.

externl: Ok, fair enough.

bernardnormier (approved):

PR #444, opened 2025-05-01
Sets .bat files to use eol=crlf and gradlew files to use eol=lf

ayoitslilith:
A simple fix in the .gitattributes file intended to fix some problems regarding line endings for scripts.


.gitattributes
```diff
@@ -4,3 +4,7 @@
 *.iml export-ignore
 .gitignore export-ignore
 .gitattributes export-ignore
+
+# Use CRLF line endings for specific file types
+*.bat text eol=crlf
+**/gradlew eol=lf
```
externl, line 10: Missing ending newline

.gitattributes
```diff
@@ -4,3 +4,7 @@
 *.iml export-ignore
 .gitignore export-ignore
 .gitattributes export-ignore
+
+# Use CRLF line endings for specific file types
```
pepone, line 8: What about: # Set line endings: CRLF for Windows scripts, LF for Unix scripts

pepone (approved):

.gitattributes
ayoitslilith: I've set it to that for now, but do you think it may be a good idea to specify that currently the only Unix scripts included by this are gradlew scripts? I can't think of any non-gradlew scripts that need this change off of the top of my head, hence the current specificity.

externl (approved):

PR #447, opened 2025-05-09
Add cpp/Ice/customError

externl:


cpp/Ice/customError/Chatbot.cpp
```diff
@@ -0,0 +1,48 @@
...
+    cout << "Dispatching greet request { name = '" << name << "' }" << endl;
+
+    if (name.empty())
+    {
+        // ObjectNotExistException is a dispatch exception with a reply status of 'ObjectNotExist'.
+        throw Ice::ObjectNotExistException{__FILE__, __LINE__};
+    }
+    else if (name == "alice")
+    {
+        // This is a bogus error since there is no authentication in this demo.
```
pepone, line 24: I would prefer if we use "dummy" instead of "bogus".

pepone (approved):

cpp/Ice/customError/Chatbot.cpp
externl: I coped it from C#, should I change it there as well?

pepone: Yes. I think we should.

bernardnormier: Make `fake`? I don't find dummy appropriate here.

cpp/Ice/customError/Greeter.ice
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+module VisitorCenter
+{
+    /// Represents the error code carried by GreeterException.
+    enum GreeterError { NameTooLong, Away, GreetingOtherVisitor };
+
+    /// Represents the custom error returned by the Greeter when it cannot create a greeting. This exception is thrown
+    /// by the implementation of the greet operation and caught by the caller.
+    exception GreeterException {
```
bernardnormier, line 12: Incorrect formatting.

bernardnormier (approved):

cpp/Ice/customError/Chatbot.cpp
pepone: Fake is fine. Or we can also write something like:

```
/ Simulate an authentication error by throwing a dispatch exception with the Unauthorized error code.
// Note: This is a demo; no real authentication logic is implemented.
```

PR #448, opened 2025-05-13
Demo addition: Java Ice cancellation

ayoitslilith:
This PR adds the cancellation demo for Ice to Java's catalogue of demos. This version of the cancellation demo does not use Ice's `InvocationCanceledException`, but instead uses `OperationInterruptedException` for the same purpose.


java/Ice/cancellation/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```shell
+./gradlew build
+```
+
+This will compile each project and install the distributions in the `build/install` subdirectory of each project.
+
+## Running the Server
+
+First, start one of the server applications:
```
bernardnormier, line 18: There is only one server in this demo.

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
```diff
@@ -0,0 +1,68 @@
...
+                GreeterPrx.createProxy(communicator, "slowGreeter:tcp -h localhost -p 4061");
+
+            // Send a request to the regular greeter and get the response.
+            String greeting = greeter.greet(System.getProperty("user.name"));
+            System.out.println(greeting);
+
+            // Create another slow greeter proxy with an invocation timeout of 4 seconds (the default invocation timeout
+            // is infinite).
+            GreeterPrx slowgreeter4s =
+                GreeterPrx.uncheckedCast(slowgreeter.ice_invocationTimeout(4));
```
bernardnormier, line 32: I believe the uncheckedCast is not needed in Java.

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
```diff
@@ -0,0 +1,68 @@
...
+
+            // Send a request to the slow greeter with the 4-second invocation timeout.
+            try {
+                greeting = slowgreeter4s.greet("alice");
+                System.out.println(greeting);
+            }   catch (InvocationTimeoutException exception) {
```
bernardnormier, line 38: formatting does not look correct

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
```diff
@@ -0,0 +1,68 @@
...
+                System.out.println("Caught InvocationTimeoutException, as expected: " + exception.getMessage());
+            }
+
+            // Send a request to the slow greeter, and cancel this request after 4 seconds.
+            Runnable greetRunnable  = () -> {
```
bernardnormier, line 43: That's too complicated. It would be nicer to make the invocation from the main thread, and find a _simple_ way to interrupt this thread after 4 seconds.

java/Ice/cancellation/README.md
```diff
@@ -0,0 +1,32 @@
+# Ice Cancellation
+
+The Cancellation demo shows how to cancel an invocation using the Java CancellationException. It also shows a related
```
bernardnormier, line 3: You meant:
shows how to cancel an invocation by interrupting the thread that waits for the response?

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
```diff
@@ -0,0 +1,68 @@
...
+                try {
+                    String greetingEnclosed = slowgreeter.greet("bob"); 
+                    System.out.println(greetingEnclosed);
+                }
+                catch (OperationInterruptedException exception) {
+                    System.out.println("Caught OperationInteruptedException, as expected.");
```
bernardnormier, line 49: two R in interrupt

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
ayoitslilith: I'm unsure what I would do otherwise here? I'm aware of a form of `createProxy`  that would accept the arguments used here.

java/Ice/cancellation/client/src/main/java/com/example/ice/cancellation/client/Client.java
ayoitslilith: The formatting regarding the
```try {
//imagine code here
} catch (ExampleException exception) {
//imagine code here
}```
is comparable to how similar try-catch statements are formatted in Ice's source code. Would you prefer if it was formatted differently?

PR #457, opened 2025-05-17
Add Java secure demo

pepone:
This demo shows how to configure Java IceSSL using the Java SSL programatic APIs.


java/Ice/secure/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) ZeroC, Inc.
+
+plugins {
+    // Register the Ice Builder plugin without applying it here.
```
bernardnormier, line 4: We should not call the Slice tools "Ice Builder plugin".

java/Ice/secure/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+    // The plugin will be applied to all subprojects in the `subprojects` block below.
+    id("com.zeroc.ice.slice-tools") version "3.8.0-+" apply false
+}
+
+subprojects {
+    // Apply the application plugin to enable Java application support in subprojects.
+    apply(plugin = "application")
+
+    // Apply the Gradle Slice Tools for Ice and Java plugin to enable Slice compilation in all subprojects.
```
bernardnormier, line 13: "Gradle Slice Tools for Ice and Java plugin" is a bit long. Why not simply the "Slice tools plugin"? We don't refer to the application plugin above as the "Gradle application for Java plugin"!

java/Ice/secure/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+    apply(plugin = "com.zeroc.ice.slice-tools")
+
+    repositories {
+        // Add the ZeroC Nightly repository for downloading Ice runtime and development artifacts.
```
bernardnormier, line 17: > for downloading Ice runtime and development artifacts.

This is not syntactically correct, and I am not fond of the expression "Ice runtime and development artifacts".

java/Ice/secure/client/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ZeroC, Inc.
+
+dependencies {
+    // Add the Ice library as an implementation dependency.
+    implementation("com.zeroc:ice:3.8.0-+")
+}
+
+sourceSets {
+    main {
+        // Add the Greeter.ice file from the parent slice directory the main source set.
```
bernardnormier, line 10: _to_  the main source set?

java/Ice/secure/client/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+        slice {
+            srcDirs("../slice")
+            // By default a Slice source set includes all Slice files in the srcDirs directories
```
bernardnormier, line 13: missing period.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,103 @@
...
+        // In a production environment, use a secure method to store and retrieve this password.
+        char[] password = "password".toCharArray();
+
+        // Create a KeyStore object to load the CA's certificate.
+        // The PKCS12 format is recommended for modern Java applications.
+        KeyStore keyStore;
+        try {
+            keyStore = KeyStore.getInstance("PKCS12");
+        } catch (KeyStoreException ex) {
+            throw new RuntimeException("Error initializing PKCS12 keystore.", ex);
```
bernardnormier, line 32: What about handling it with:
```java
} catch (KeyStoreException ex) {
    // Should never happen.
    ex.printStackTrace();
    System.exit(1);
}
```

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        // Load the CA certificate from the keystore file into the KeyStore object.
+        String keyStorePath = "../../../../certs/cacert.p12";
+        try (var input = new FileInputStream(keyStorePath)) {
+            keyStore.load(input, password);
+        } catch (Exception ex) {
```
bernardnormier, line 39: Same here, except I'd prefer a more specific catch like
```java
catch (IOException | NoSuchAlgorithmException | CertificateException ex) {
```

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        // Create a TrustManagerFactory to manage the trusted certificates.
+        // This will be used by the SSLContext to verify that the server's certificate was signed by our demo
+        // custom CA.
+        TrustManagerFactory trustManagerFactory;
+        try {
+            trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            trustManagerFactory.init(keyStore);
+        } catch (Exception ex) {
```
bernardnormier, line 50: Ditto

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,103 @@
...
+            // "stringified proxy" with the address of the target object.
+            // If you run the server on a different computer, replace localhost in the string below with the server's
+            // hostname or IP address.
+            GreeterPrx greeter = GreeterPrx.createProxy(communicator, "greeter:ssl -h localhost -p 4061");
+
+            // Send a request to the remote object and get the response.
+            String greeting = greeter.greet(System.getProperty("user.name"));
+            System.out.println(greeting);
+
+            // Send another request to the remote object, this time with greetAsync.
```
bernardnormier, line 89: What's the point of showing the async API in this demo about SSL?

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,103 @@
...
+
+        // Create an SSLContext using the TLS protocol.
+        SSLContext sslContext;
+        try {
+            sslContext = SSLContext.getInstance("TLS");
+        } catch (NoSuchAlgorithmException ex) {
+            throw new RuntimeException("Error initializing TLS protocol.", ex);
+        }
+
+        // Initialize the SSLContext with the trust managers from the TrustManagerFactory.
```
bernardnormier, line 62: I find all these catch blocks over the top. Surely we can merge some of them.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
pepone: The compiler complains that keystore might be not initialized, it understand returns and throws, but doesn't seems to understand System.exit doesn't return.

java/Ice/secure/build.gradle.kts
pepone: How about `// Add the ZeroC Nightly maven repository, required to use the Ice nightly builds.`

java/Ice/context/client/build.gradle.kts
```diff
@@ -7,12 +7,9 @@ dependencies {
 
 sourceSets {
     main {
-        // Add the Greeter.ice file from the parent slice directory the main source set.
+        // Add the Greeter.ice file from the parent slice directory to the main source set.
         slice {
             srcDirs("../slice")
-            // By default a Slice source set includes all Slice files in the srcDirs directories
-            // Here we override the default behavior by specifying the list of Slice files to include.
-            setIncludes(listOf("Greeter.ice"))
```
pepone, line 15: This is required in Ice/greeter because we have both Greeter.ice and GreeterAMD.ice and only want to include one. But not required here or in the new secure demo.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
bernardnormier: Ok. Maybe the solution is to keep this RunException wrapping for all these checked exceptions that never occur, and also merge all these try/catch block for better readability?

java/Glacier2/greeter/build.gradle.kts
```diff
@@ -10,11 +10,11 @@ subprojects {
     // Apply the application plugin to enable Java application support in subprojects.
     apply(plugin = "application")
 
-    // Apply the Gradle Slice Tools for Ice and Java plugin to enable Slice compilation in all subprojects.
+    // Apply the Slice tools plugin to enable Slice compilation in all subprojects.
     apply(plugin = "com.zeroc.ice.slice-tools")
 
     repositories {
-        // Add the ZeroC Nightly repository for downloading Ice runtime and development artifacts.
+        // Add the ZeroC Nightly maven repository, required to use the Ice nightly builds.
```
bernardnormier, line 17: What about:
```
// This demo uses the latest Ice nightly build published in ZeroC's maven-nightly repository.
```

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,73 @@
...
+                TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            trustManagerFactory.init(keyStore);
+            sslContext.init(null, trustManagerFactory.getTrustManagers(), null);
+        } catch (
+            CertificateException|
+            IOException|
+            KeyManagementException|
+            KeyStoreException|
+            NoSuchAlgorithmException ex) {
+            throw new RuntimeException("SSL initialization error.", ex);
```
bernardnormier, line 48: I prefer this larger catch block. I would also add a comment like:

```java
// Should never happen in this demo.
```

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
pepone: Yes. That is what I end-up doing.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManagerFactory;
+
+class Client {
+    public static void main(String[] args) {
+
+        // The password for the PKCS12 keystore, hard-coded for simplicity.
+        // In a production environment, use a secure method to store and retrieve this password.
+        char[] password = "password".toCharArray();
```
bernardnormier, line 25: I would move this into the sslContext initialization block below.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
bernardnormier: Maybe putting this code in a helper function, `createSslContext`, would improve readability?

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,73 @@
...
+            KeyManagementException|
+            KeyStoreException|
+            NoSuchAlgorithmException ex) {
+            throw new RuntimeException("SSL initialization error.", ex);
+        }
+
+        var initData = new com.zeroc.Ice.InitializationData();
+        initData.properties = new com.zeroc.Ice.Properties(args);
+        
+        // The clientSSLEngineFactory lambda provides the communicator with an SSLEngine that uses our custom
```
bernardnormier: I find `custom` misleading here. It's a plain Java SSLContext that we've configured to our liking.

What about:
```java
// Each time the communicator creates a new outgoing secure connection, it uses the 
// `clientSSLEngineFactory` to create a new SSLEngine for this connection. Here, we create these 
// SSLEngine using our carefully crafted SSLContext.
```

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,73 @@
...
+        // SSLContext for each outgoing secure connection.
+        initData.clientSSLEngineFactory =
+            (String peerHost, int peerPort) -> sslContext.createSSLEngine(peerHost, peerPort);
+
+        // Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+        try (Communicator communicator = Util.initialize(initData)) {
+            // GreeterPrx is a class generated by the Slice compiler. We create a proxy from a communicator and a
+            // "stringified proxy" with the address of the target object.
+            // If you run the server on a different computer, replace localhost in the string below with the server's
```
bernardnormier, line 62: We should keep this comment about replacing localhost by some other hostname/IP address only if it actually works with the demo. With SSL and certificates, it's not obvious to me that it works.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
bernardnormier: Is the formatting correct?

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -0,0 +1,82 @@
...
+            // In a production environment, use a secure method to store and retrieve this password.
+            char[] password = "password".toCharArray();
+
+            try (var input = new FileInputStream(keyStorePath)) {
+                keyStore.load(input, password);
+            }
+            TrustManagerFactory trustManagerFactory =
+                TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            trustManagerFactory.init(keyStore);
+            sslContext.init(null, trustManagerFactory.getTrustManagers(), null);
```
bernardnormier, line 70: Why not simply return sslContext at the end of the try/catch block?

bernardnormier (approved):
Looks good.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
pepone: Fixed.

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
pepone: Removed.

externl (approved):

externl:
I'm wondering if we should have a comment mentioning that this is showing one-way TLS, not mTLS.

PR #463, opened 2025-05-21
Demo addition: Java Ice middleware

ayoitslilith:
This PR adds the middleware demo to Java's catalog of Ice demos.


java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
...
+import com.zeroc.Ice.IncomingRequest;
+import com.zeroc.Ice.Object;
+import com.zeroc.Ice.OutgoingResponse;
+import com.zeroc.Ice.UserException;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * The AuthorizationMiddleware intercepts incoming requests and checks the token entry in their contexts.
+ */
+public class AuthorizationMiddleware implements Object {
```
bernardnormier, line 15: I would make this class package private, like in C#.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.example.ice.middleware.server;
+
+import com.zeroc.Ice.DispatchException;
+import com.zeroc.Ice.IncomingRequest;
+import com.zeroc.Ice.Object;
```
bernardnormier, line 7: I feel it would be clearer to always qualify com.zeroc.Ice.Object to avoid confusion with java.lang.Object.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
...
+    private final Object _next;
+    private final String _validToken;
+
+    public CompletionStage<OutgoingResponse> dispatch(IncomingRequest request) {
+        String token = request.current.ctx.get("token");
+
+        // Check if the request has a valid token in its context.
+        if (token == null || !token.equals(_validToken)) {
+            System.out.println("Rejecting request with invalid token " + token);
+            throw new DispatchException(9, "Invalid token " + token);
```
bernardnormier, line 25: What is this 9?
We should naturally use the enumerator.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        System.out.println("Accepting request with token " + token);
+        
+        // Continue the dispatch pipeline.
+        CompletionStage<OutgoingResponse> response = null;
```
bernardnormier, line 31: All this code should be simply:
```java
// Continue the dispatch pipeline.
return _next.dispatch(request);
```

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.CompletionStage;
+
+/**
+ * The AuthorizationMiddleware intercepts incoming requests and checks the token entry in their contexts.
+ */
+public class AuthorizationMiddleware implements Object {
+    private final Object _next;
+    private final String _validToken;
+
+    public CompletionStage<OutgoingResponse> dispatch(IncomingRequest request) {
```
bernardnormier, line 19: missing exception specification

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,50 @@
...
+            response = _next.dispatch(request);
+        }
+        catch (UserException exception)
+        {
+            System.out.println("Caught unexpected UserException " + exception.getMessage());
+        }
+        return response;
+    }
+
+    // Constructs an AuthorizationMiddleware.
```
bernardnormier, line 43: This is should be a properly formatted doc-comment.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
ayoitslilith: I don't believe the enumerator works here. The `9` corresponds with the `Unauthorized` enum in ReplyStatus.java, but neither ReplySttatus.java nor DispatchException.java have a function to potentially use the enum as an integer during the creation of a `DispatchException`.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
ayoitslilith: Java does not like it when `UserException` isn't caught here - attempting to directly use `return _next.dispatch(request);` will result in a compilation error.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
ayoitslilith: Private isn't permitted here - leaving it modifier-less may give a more similar effect to what's going on in C#, however.

bernardnormier: Yes, it's called "package private" in Java.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+        if (token == null || !token.equals(_validToken)) {
+            System.out.println("Rejecting request with invalid token " + token);
+            throw new DispatchException(ReplyStatus.Unauthorized.value(), "Invalid token " + token);
+        }
+
+        System.out.println("Accepting request with token " + token);
+        
+        // Continue the dispatch pipeline.
+        return _next.dispatch(request);
+
```
bernardnormier, line 32: extra blank line

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     *  Constructs an AuthorizationMiddleware.
+     * @param next The next Ice.Object in the dispatch pipeline.
+     * @param validToken The value to check against.
```
InsertCreativityHere, line 38: A couple small things.
There's an extra space in front of `Constructs`, and it's customary not for `@param` to:
1) not end with a period 2) start with a lowercase letter. So, they aren't supposed to be "sentences".
Although, I only just learned the 2nd point myself yesterday, and we appear to violate it in all of our source code.
So, I don't blame you for having it here!
```suggestion
     * Constructs an AuthorizationMiddleware.
     * @param next the next Ice.Object in the dispatch pipeline
     * @param validToken the value to check against
```

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+class AuthorizationMiddleware implements com.zeroc.Ice.Object {
+    private final com.zeroc.Ice.Object _next;
+    private final String _validToken;
+
+    public CompletionStage<OutgoingResponse> dispatch(IncomingRequest request) throws DispatchException, UserException {
+        String token = request.current.ctx.get("token");
+
+        // Check if the request has a valid token in its context.
+        if (token == null || !token.equals(_validToken)) {
+            System.out.println("Rejecting request with invalid token " + token);
```
InsertCreativityHere, line 24: To be fully consistent with the C# version, we should put the tokens in single quotes.
Also, I just think it's a good idea to wrap variables in quotes when printing them, so it's clearly delineated from the surrounding text.
```suggestion
            System.out.println("Rejecting request with invalid token '" + token + "'");
```

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+            throw new DispatchException(ReplyStatus.Unauthorized.value(), "Invalid token " + token);
```
InsertCreativityHere, line 25: Oh, same here I guess:
```suggestion
            throw new DispatchException(ReplyStatus.Unauthorized.value(), "Invalid token '" + token + "'");
```

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        System.out.println("Accepting request with token " + token);
```
InsertCreativityHere, line 28: Same here:
```suggestion
        System.out.println("Accepting request with token '" + token + "'");
```

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
```diff
@@ -0,0 +1,44 @@
...
+        
+        // Continue the dispatch pipeline.
+        return _next.dispatch(request);
+
+    }
```
InsertCreativityHere, line 33: Seems like an extra newline here:
```suggestion
        return _next.dispatch(request);
    }
```

InsertCreativityHere (approved):
Looks good to me, just a few minor things.

java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java
ayoitslilith: Thanks for the clarification - the source code was what I referenced for this, so the punctuation was definitely a point of confusion for me.

bernardnormier (approved):

PR #465, opened 2025-05-21
Fix GreeterService::stop

### LINKED ISSUE #464: Incorrect GreeterService::stop in C++
The C++ IceBox/greeter demo should be like the C# demo:

```cs
 public void stop()
    {
        Console.WriteLine($"{_greeterName} is shutting down...");

        // We destroy the object adapter in stop() in case the service is started, stopped and then restarted again
        // programmatically or using an admin tool.
        // The general rule is stop should cleanup all resources created by start.
        Debug.Assert(_adapter is not null);
        _adapter.destroy();
        _adapter = null;
    }
```

bernardnormier:
This PR fixes #464.

### DIFF
--- a/cpp/IceGrid/icebox/GreeterService.h
+++ b/cpp/IceGrid/icebox/GreeterService.h
@@ -23,7 +23,7 @@ namespace Service
         std::string _greeterName{"unknown"};
 
         // The object adapter created by this service.
-        Ice::ObjectAdapterPtr _adapter{nullptr};
+        Ice::ObjectAdapterPtr _adapter;
     };
 }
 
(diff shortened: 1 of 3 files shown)


cpp/IceGrid/icebox/GreeterService.h
```diff
@@ -23,7 +23,7 @@ namespace Service
         std::string _greeterName{"unknown"};
 
         // The object adapter created by this service.
-        Ice::ObjectAdapterPtr _adapter{nullptr};
```
bernardnormier, line 26: We typically don't/shouldn't initialize to the default value.

externl (approved):

PR #466, opened 2025-05-22
Update filesystem demo (Swift)

bernardnormier:


swift/Ice/filesystem/Package.swift
```diff
@@ -0,0 +1,26 @@
+// swift-tools-version: 5.9
+
+import Foundation
+import PackageDescription
+
+let package = Package(
+    name: "filesystem",
+    platforms: [
+        .macOS(.v14)
```
bernardnormier, line 9: This comes from the greeter demo model. It's not clear to me why we're targeting macos 14.

pepone (approved):

PR #477, opened 2025-05-27
Improve Java Greeter

bernardnormier:
This PR applies various updates to the Greeter demo. Once merged, we should apply the same changes to all other Java demos.


InsertCreativityHere (approved):
I am always in favor of more consistency!

PR #480, opened 2025-05-28
Add new Ice/forwarder demo (Swift)

bernardnormier:


pepone (approved):

csharp/Ice/Forwarder/Client/Program.cs
```diff
@@ -6,7 +6,7 @@
 // Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
 await using Ice.Communicator communicator = Ice.Util.initialize(ref args);
 
-// We create a Greeter proxy for a Greeter object in the Forwarder server (port 10000).
+// We create a Greeter proxy for a Greeter object in the Forwarding server (port 10000).
```
externl, line 9: I think this should be lowercase or kept as `Forwarder`.

swift/Ice/forwarder/Sources/Client/main.swift
```diff
@@ -0,0 +1,23 @@
...
+// Create an Ice communicator. We'll use this communicator to create proxies and manage outgoing connections.
+var args = CommandLine.arguments
+let communicator = try Ice.initialize(&args)
+
+// Destroy the communicator when the program exits.
+defer {
+    communicator.destroy()
+}
+
+// We create a Greeter proxy for a Greeter object in the Forwarding server (port 10000).
```
externl, line 15: Ditto

externl (approved):

csharp/Ice/Forwarder/Client/Program.cs
bernardnormier: The name of the server is Forwarding Server. The demo is called Forwarder, and the servant is called Forwarder.

externl: I missed that it was called "Forwarding Server"

PR #484, opened 2025-05-29
Upgrade Package.swift of Swift demos

bernardnormier:


bernardnormier:
Various demos don't compile with Swift 6.0 strict concurrency checks. Needs to fix Ice first.

PR #506, opened 2025-06-02
Cleanup and Synchronize Java Demo Build Files

InsertCreativityHere:
1) Each demo has a "root-level" `build.gradle.kts` file, and another `build.gradle.kts` in each subproject (`client` and `server` usually). This "root-level" file pushes configuration into the others via a `subproject` block. Nowadays this is [frowned upon](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration). I removed most of the root-level build files, and their `subproject` blocks.

2) We're using an outdated approach to apply plugins. With the Kotlin DSL, [plugins should always be applied declaratively](https://docs.gradle.org/current/userguide/best_practices_general.html#use_the_plugins_block) inside a `plugins {}` block. Instead of programatically by calling the `apply(plugin = ...)` function. We were actually using a mix of the old/new approaches. I switched all the old `apply` calls over.

3) We're using an outdated way to declare our repositories. And it's split between 2 separate files. We declare our plugin repositories in `settings.gradle.kts` (good), but our dependency repositories in `build.gradle.kts` (bad). [Repositories should always be set in](https://docs.gradle.org/current/userguide/best_practices_dependencies.html#set_up_repositories_in_settings) `settings`. And we should be using the new `dependencyResolutionManagement` instead of the old `repositories` block.

4) We still had a mix of Ice version numbers. Some of them were changed in a previous PR, but not all. I'm switching them all to use `3.8.+`. This will use the nightly packages, until we release `3.8.0`, at which point it will start using those instead (stable packages always 'outrank' nightlies in priority).

### DIFF
--- a/java/Ice/forwarder/forwardingserver/build.gradle.kts
+++ b/java/Ice/forwarder/forwardingserver/build.gradle.kts
@@ -3,9 +3,6 @@
 plugins {
     // Apply the application plugin to tell gradle this is a runnable Java application.
     id("application")
-
-    // Apply the Slice-tools plugin to enable Slice compilation.
-    id("com.zeroc.ice.slice-tools") version "3.8.+"
 }
 
 dependencies {


java/Ice/config/settings.gradle.kts
```diff
@@ -3,13 +3,19 @@
...
-rootProject.name = "greeter"
+dependencyResolutionManagement {
+    repositories {
+        // This demo uses the latest Ice nightly build published in ZeroC's maven-nightly repository.
+        maven("https://download.zeroc.com/nexus/repository/maven-nightly/")
+        mavenCentral()
+    }
+}
+
+rootProject.name = "config"
```
InsertCreativityHere, line 19: Some of the project names were copy/pasted without ever being updated.
This is the `config` demo`. Not the `greeter` demo, from which it was copied.

java/Ice/middleware/settings.gradle.kts
```diff
@@ -3,13 +3,19 @@
...
-rootProject.name = "greeter"
+dependencyResolutionManagement {
+    repositories {
+        // This demo uses the latest Ice nightly build published in ZeroC's maven-nightly repository.
+        maven("https://download.zeroc.com/nexus/repository/maven-nightly/")
+        mavenCentral()
+    }
+}
+
+rootProject.name = "middleware"
```
InsertCreativityHere, line 19: Another bad name. Should be the `middleware` demo.

java/Ice/secure/settings.gradle.kts
```diff
@@ -3,13 +3,19 @@
...
-rootProject.name = "context"
+dependencyResolutionManagement {
+    repositories {
+        // This demo uses the latest Ice nightly build published in ZeroC's maven-nightly repository.
+        maven("https://download.zeroc.com/nexus/repository/maven-nightly/")
+        mavenCentral()
+    }
+}
+
+rootProject.name = "secure"
```
InsertCreativityHere, line 19: Should be the `secure` demo, not `context`.

java/Ice/forwarder/forwardingserver/build.gradle.kts
```diff
@@ -1,8 +1,16 @@
 // Copyright (c) ZeroC, Inc.
 
+plugins {
+    // Apply the application plugin to tell gradle this is a runnable Java application.
+    id("application")
+
+    // Apply the Slice-tools plugin to enable Slice compilation.
+    id("com.zeroc.ice.slice-tools") version "3.8.+"
```
pepone, line 8: Not needed for the forwarding server.

java/Ice/forwarder/build.gradle.kts
```diff
@@ -1,24 +1,6 @@
...
-    // Apply the Slice tools plugin to enable Slice compilation in all subprojects.
-    apply(plugin = "com.zeroc.ice.slice-tools")
-
-    repositories {
-        // This demo uses the latest Ice nightly build published in ZeroC's maven-nightly repository.
-        maven("https://download.zeroc.com/nexus/repository/maven-nightly/")
-        mavenCentral()
-    }
-
     tasks.withType<JavaCompile>().configureEach {
```
pepone, line 4: Do you plan to rework the JavaCompile settings? Seems contradictory to move everything out and keep it just for this.

pepone (approved):

java/Ice/forwarder/build.gradle.kts
InsertCreativityHere: Yep, these files are going away in my next PR.
Just waiting to move these last few settings until I also add support for our formatter/linter.
Since they're going to end up in the same shared build script.

bernardnormier (approved):

PR #509, opened 2025-06-03
Add `checkstyle` and `rewrite` Linting to the Java Demos / More Gradle Cleanup

InsertCreativityHere:
This PR:
- Adds and configures the `checkstyle` and `OpenRewrite` plugins to format/lint the Java demos. It's using the same configuration as `ice/java` (even though this config needs to be cleaned up IMO).
- Adds a github workflow to run `check` and `rewriteDryRun` in CI.
- Fixes all the mistakes they caught in the generated code.

----

I created a so-called "convention plugin" named `linting-config` which is applied to all our projects just like any other plugin.
This is the [preferred way to share build-logic](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#using_a_composite_build_named_build_logic) between multiple projects nowadays. It lives in `java/config/build-logic`.


java/Glacier2/greeter/settings.gradle.kts
```diff
@@ -19,3 +19,5 @@ dependencyResolutionManagement {
 rootProject.name = "greeter"
 include("client")
 include("server")
+
+includeBuild("../../config/build-logic")
```
InsertCreativityHere, line 23: Tells Gradle that it needs to build our local convention-plugin, before it can build the rest of this project.

.github/workflows/java.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Run checkstyle formatter
+        timeout-minutes: 20
+        working-directory: java
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          Get-ChildItem -Recurse -Depth 2 -Filter gradlew | ForEach-Object {
+            Push-Location $_.DirectoryName
+            Write-Host "Building in $($_.DirectoryName)..."
+            $gradlew = if ($IsWindows) { "./gradlew.bat" } else { "./gradlew" }
```
pepone, line 32: The conditional Gradle is not required here, this only runs in Ubuntu.

pepone (approved):

java/Glacier2/greeter/client/build.gradle.kts
```diff
@@ -6,6 +6,9 @@ plugins {
 
     // Apply the Slice-tools plugin to enable Slice compilation.
     id("com.zeroc.ice.slice-tools") version "3.8.+"
+
+    // Pull in our local 'convention plugin' to enable linting.
+    id("linting-config")
```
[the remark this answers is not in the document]
InsertCreativityHere: It's worth noting that you can have multiple convention plugins like:
```
build-logic/
    build.gradle.kts
    src/main/kotlin/
        linting-config.gradle.kts
        assertion-stuff.gradle.kts
        slice-setup.gradle.kts
```
We just happen to only have one. So, I'd prefer to have more focused names, instead of a single vaguely named monolith. How would you feel about `zeroc-linting`? I agree having `config` in a plugin name isn't great.

java/Glacier2/greeter/settings.gradle.kts
InsertCreativityHere: We can do anything of course, but `config` is the expected name for checkstyle and rewrite (when it's not in the project root at least). So, if we want to still follow conventions, we could do two directories instead:
```
build-common/
    all-the-plugin-stuff/
        ...
    config/
        rewrite.yml
        checkstyle
            checkstyle.xml
            suppressions.xml
```
Fine with you?

InsertCreativityHere: Or maybe just this, to eliminate an extra layer of directories to go through. And so that checkstyle and rewrite are still in their expected places (since the "project root" of the demos is logically the `java` folder):
```
build-common/
    ... all-the-plugin-stuff/
config/
    rewrite.yml
    checkstyle/
        checkstyle.xml
        suppressions.xml
```

java/config/build-logic/src/main/kotlin/linting-config.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    // 'rewrite' uses a modular plug-in system for linting. We pull in the recipes we want to use.
+    rewrite(
+        "org.openrewrite.recipe:rewrite-static-analysis:2.5.1"
+    )
+}
+
+checkstyle {
+    toolVersion = "10.21.4"
+
+    configDirectory.set(file("$rootDir/../../config/checkstyle/"))
```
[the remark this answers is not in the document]
InsertCreativityHere: `rootDir` is the directory containing the `settings.gradle` file.
In this case these would be things like `java/Ice/config/`, or `java/Glacier2/greeter`.
Since each demo is it's own gradle project.

java/Glacier2/greeter/client/build.gradle.kts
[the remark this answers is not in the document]
InsertCreativityHere: I renamed it to `zeroc-linting`.

PR #517, opened 2025-06-05
Add Android Greeter demo

### LINKED ISSUE #515: Create Android Version of Ice/greeter
We want to build an Android version of the existing java/Ice/greeter/client demo.

## Requirements

The application should provide the following UI elements and behavior:
- An input field to enter the IP address of the target server.
  - Default value: 10.0.2.2 (to connect to the host machine from an Android emulator)
- An input field to enter a name.
  - Default value: "Android"
- A button to send the greet request.
  - Upon receiving a response, the application should display the greeting message on the screen.
  
## Implementation Notes
- Use the greeter.greetAsync method to send the request.
  - The call must be non-blocking to ensure the UI remains responsive.
- Handle errors gracefully and display an error message on the UI if the request fails.
- The Ice communicator must be:
  - Initialized when the application starts.
  - Destroyed when the application is finished (typically in onTerminate() or appropriate lifecycle method).

pepone:
Fix #515


java/Ice/greeter-android/app/src/main/java/com/example/ice/androidgreeter/App.kt
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.example.ice.androidgreeter
```
bernardnormier, line 3: It would be nice for the demo name and package name to be consistent: either greeter android or android greeter, not demo=greeter-android and package=androidgreeter.

I prefer "Android Greeter".

java/Ice/greeter-android/app/src/main/java/com/example/ice/androidgreeter/MainActivity.kt
```diff
@@ -0,0 +1,83 @@
+package com.example.ice.androidgreeter
```
bernardnormier, line 1: Missing copyright line

java/Ice/greeter-android/app/src/main/java/com/example/ice/androidgreeter/MainActivity.kt
```diff
@@ -0,0 +1,83 @@
...
+            greetButton.isEnabled = false
+
+            // Launch a coroutine tied to the activity's lifecycle
+            lifecycleScope.launch {
+                try {
+                    // Create a proxy to the remote Greeter object.
+                    // The proxy is recreated for each request to ensure the latest IP address is used.
+                    val greeter = GreeterPrx.createProxy(communicator, "greeter:tcp -h $ip -p 4061")
+
+                    // Send the greetAsync request on a background thread and wait for the response
```
bernardnormier, line 62: Is this correct?

The Ice async API does not block the calling thread so it should not be necessary to send this request "on a background thread".

java/Ice/android-greeter/app/src/main/java/com/example/ice/androidgreeter/App.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.zeroc.Ice.Util
+
+/**
+ * A custom [Application] class that initializes and owns the Ice [Communicator].
+ *
+ * The communicator is initialized once at application startup and is made available
+ * through [getCommunicator]. If initialization fails, the application will crash immediately.
+ */
+class App : Application() {
+    private var communicator: Communicator? = null
```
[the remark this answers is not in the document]
pepone: it is not, application remains in the background until is killed. 

The Android app life cycle is a bit complex I think it is not worth to enter into these details for a simple greeter.
https://developer.android.com/guide/components/activities/activity-lifecycle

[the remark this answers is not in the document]
externl: Can't we use https://developer.android.com/guide/components/activities/activity-lifecycle#ondestroy?

PR #521, opened 2025-06-06
Convert class to structs in Swift demos

bernardnormier:
This PR converts most classes to struct in the Swift demos. Using structs is more common in Swift.


bernardnormier:
> Looks good.
> 
> What is the rule here? We prefer structs for Ice services, unless you need inheritance?

Typically:
 - use structs for stateless servants
 - use structs for servants that delegate to some other struct/class/actor
 - use actors for servants with mutable state
 
And yes, use classes if you want to reuse a servant implementation via inheritance. This should be ultra-rare as implementation-reuse through inheritance is not a common Swift pattern.

PR #523, opened 2025-06-07
Use latest Xcode

pepone:


bernardnormier:
#484 already upgrades all the Package.swift.

PR #524, opened 2025-06-07
Rework Swift Ice/filesystem server to use actor and delegation

bernardnormier:
This PR reworks the Swift Ice/filesystem server to use actors (for MFile and MDirectory) and delegation for the MNode reuse.

This style is definitely more consistent with Swift current practices.

We could provide another demo that showcases how to reuse a servant implementation using inheritance.

However:
 - this demo would be about showing that this layout is possible, but not recommended
 - the focus of the demos is not quirky things you can do with Ice

So I'd rather not include such a "implementation reuse via inheritance" demo in Swift.


swift/Ice/filesystem/Sources/Server/MDirectory.swift
```diff
@@ -3,24 +3,27 @@
 import Ice
 
 /// Provides an in-memory implementation of the Slice interface Directory.
-// This class reuses the implementation of Node provided by MNode.
-final class MDirectory: MNode, Directory {
+/// We use an actor since this servant has mutable state (contents).
+actor MDirectory: Directory {
+    private let node: MNode  // reuse the MNode implementation
```
externl, line 8: Maybe we should use the term composition?

externl (approved):

pepone (approved):

PR #526, opened 2025-06-10
Fix ca cert listed in config files

externl:
The cert was rename. @pepone is there anything I may have missed?


pepone (approved):
Looks good

PR #532, opened 2025-06-18
Update MATLAB demos

bernardnormier:


matlab/Ice/customError/Greeter.ice
```diff
@@ -13,10 +13,11 @@ module VisitorCenter
     exception GreeterException
     {
         /// Describes the exception in a human-readable way.
-        ["matlab:identifier:errorMessage"] // We can't use message as it conflicts with MException.message.
+        ["matlab:identifier:Message"] // We prefer Pascal case for MATLAB properties.
```
pepone, line 16: Does MATLAB use Pascal case?

PR #540, opened 2025-06-26
Update nightly instructions

pepone:
Updates the build prerequisites to link NIGHTLY instructions in zeroc-ice/ice

### DIFF
--- a/matlab/README.md
+++ b/matlab/README.md
@@ -4,19 +4,27 @@ This directory contains example Ice applications written in MATLAB. Each example
 that demonstrates a specific feature or programming technique. The corresponding server is provided by the C++, C#,
 Java, and Python demos.
 
+## Building Requirements
+
+1. **MATLAB R2024a**
+   [MATLAB](https://www.mathworks.com/products/matlab.html).
+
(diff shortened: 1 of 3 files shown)


matlab/README.md
```diff
@@ -4,19 +4,27 @@ This directory contains example Ice applications written in MATLAB. Each example
 that demonstrates a specific feature or programming technique. The corresponding server is provided by the C++, C#,
 Java, and Python demos.
 
+## Building Requirements
+
+1. **MATLAB R2024a**
```
bernardnormier, line 9: or later?

I shouldn't require an exact version when MathWorks release a new version every 6 months.

We should also specify "for Windows or Linux", since we don't support MATLAB on macos.

bernardnormier (approved):

externl (approved):

matlab/README.md
pepone: The toolbox links against a specific version of MATLAB, currently 2024a, in the pass we built toolboxes for each release. I'm not sure if is ok to link with one version of MATLAB and run in another version.

PR #548, opened 2025-07-28
Added `.gitignore`s to Each of the Java Demos

InsertCreativityHere:
This PR adds a `.gitignore` to each of the Java demos, now that the top-level `.gitignore` has been removed.

I also added one to the `build-logic` folder, where gradlew compiles and imports our build plugin to run the formatter.

### DIFF
--- a/java/Glacier2/greeter/.gitignore
+++ b/java/Glacier2/greeter/.gitignore
@@ -1,0 +1,5 @@
+bin
+build
+.gradle
+
+.DS_Store
--- a/java/Ice/cancellation/.gitignore
+++ b/java/Ice/cancellation/.gitignore
@@ -1,0 +1,5 @@
+bin
+build
+.gradle
+
+.DS_Store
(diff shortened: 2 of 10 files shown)


InsertCreativityHere:
Note that the Ice/android-greeter demo already had it's own `.gitignore` which I just left alone.

java/Glacier2/greeter/.gitignore
```diff
@@ -0,0 +1,5 @@
+bin
+build
+.gradle
+
+.DS_Store
```
bernardnormier, line 5: I would not include .DS_Store. Or if we do, it would be in the top-level .gitignore.

> A .DS_Store file is a hidden file created by macOS Finder to store custom attributes and metadata about a folder, such as icon positions, view settings, and Spotlight comments. These files are automatically generated when you open a folder with Finder, even on remote file systems. While primarily used by Finder, they can cause issues with backups, version control systems, and folder archives. 

Try:
```
find . -name .gitignore | xargs fgrep DS_Store
```

bernardnormier (approved):

PR #550, opened 2025-07-30
Update csharp/Ice/Filesystem to show implementation reuse

bernardnormier:
We use the same technique in the corresponding Swift demo.


InsertCreativityHere (approved):
Looks good to me!

PR #551, opened 2025-07-30
Add 'filesystem' Demo for Java

InsertCreativityHere:
This PR adds the 'Ice/filesystem' to Java, based off of the C# and C++ versions of this demo.
It has 'implementation re-use' via inheritance, the expected pattern in Java.


java/Ice/filesystem/server/src/main/java/com/example/ice/filesystem/server/MNode.java
```diff
@@ -0,0 +1,28 @@
...
+package com.example.ice.filesystem.server;
+
+import com.example.filesystem.Node;
+
+import com.zeroc.Ice.Current;
+
+/**
+ * Provides an in-memory implementation of the Slice interface Node.
+ */
+public class MNode implements Node {
```
bernardnormier, line 12: MNode etc. should be package-private, like Chatbot in greeter.

java/Ice/filesystem/server/src/main/java/com/example/ice/filesystem/server/MNode.java
```diff
@@ -0,0 +1,28 @@
...
+    private final String _name;
+
+    /**
+     * Constructs an {@code MNode}.
```
bernardnormier, line 16: a MNode?

Same for MFile and MDirectory.

java/Ice/filesystem/server/src/main/java/com/example/ice/filesystem/server/MDirectory.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.example.ice.filesystem.server;
+
+import com.example.filesystem.Directory;
+import com.example.filesystem.NodePrx;
+
+import com.zeroc.Ice.Current;
+
+import java.util.ArrayList;
+import java.util.List;
```
bernardnormier, line 11: What is this for?

bernardnormier (approved):

java/Ice/filesystem/server/src/main/java/com/example/ice/filesystem/server/MNode.java
InsertCreativityHere: 'an MNode' is correct here.
Even though it starts with the letter 'M', it's pronounced as an initial ('em') and not the sound 'm' normally makes.
And since the pronunciation starts with a vowel, we should use 'an' instead of 'a'.

I'll update the other demos accordingly. Didn't realize they were still using 'a'.

PR #568, opened 2025-09-04
Revert "Rename 'Filesystem' Demo to 'Inheritance' (#566)"

bernardnormier:
This reverts commit a01dd2ed623416f37c1562db411ac1f8e6ff0a9b.

This renaming was not correct. The module name should remain as Filesystem. Just like we use module VisitorCenter for the greeter, config (etc.) demos, the module ClearSky for various demos (none named ClearSky), etc.


InsertCreativityHere (approved):
I assume that you're going to re-fix the names in a separate PR.

PR #572, opened 2025-09-10
Add missing db directories for IceGrid and IceStorm

### LINKED ISSUE #570: Can't run C++ IceGrid demos
```
greeter % icegridregistry --Ice.Config=config.registry
!! 09/04/25 17:24:31.144 icegridregistry: error: property 'IceGrid.Registry.LMDB.Path' is set to an invalid path: db/registry
```

Like in the csharp demos, the directories must be provided with the demos.

bernardnormier:
Fixes #570.


externl (commented):
I think having the * in these files makes sense.

PR #577, opened 2025-09-15
Add Ice/bidir Java demo

pepone:
This PR ports the Ice/bidir demo to Java

### DIFF
--- a/java/Ice/bidir/util/src/main/java/com/example/ice/bidir/util/Time.java
+++ b/java/Ice/bidir/util/src/main/java/com/example/ice/bidir/util/Time.java
@@ -29,4 +29,7 @@ public class Time {
         long epochMilli = (timeStamp - MILLIS_BEFORE_EPOCH) / TICKS_PER_MILLISECOND;
         return java.time.Instant.ofEpochMilli(epochMilli).atZone(java.time.ZoneOffset.UTC);
     }
+
+    private Time() {
+    }
 }


java/Ice/bidir/README.md
```diff
@@ -0,0 +1,42 @@
+# Ice Bidir
+
+The Bidir demo illustrates how to send requests "the other way around", from a server to a client, by reusing the
+connection established by the client to the server.
+
+This demo is very similar to the [Callback][1] demo: with the Callback demo, the server opens a connection to the
```
bernardnormier, line 6: Missing `[1]`.

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
```diff
@@ -0,0 +1,66 @@
...
+import com.example.earlyriser.AlarmClockPrx;
+import com.example.earlyriser.ButtonPressed;
+import com.example.earlyriser.WakeUpService;
+
+import com.zeroc.Ice.Connection;
+import com.zeroc.Ice.Current;
+import com.zeroc.Ice.FeatureNotSupportedException;
+import com.zeroc.Ice.Identity;
+
+/// <summary>BidirWakeUpService is an Ice servant that implements Slice interface WakeUpService.</summary>
```
InsertCreativityHere, line 20: More C# comments. Javadoc should use `/** ... */` syntax.
(But fun-fact though, the latest Java is adding support for `///` style comments)

java/Ice/bidir/client/src/main/java/com/example/ice/bidir/client/Client.java
```diff
@@ -0,0 +1,61 @@
...
+            // Schedule a wake-up call in 5 seconds. This call establishes the connection to the server; incoming
+            // requests over this connection are handled by the communicator's default object adapter.
+            long millis = ZonedDateTime.now(ZoneOffset.UTC)
+                           .plusSeconds(10)
+                           .toInstant()
+                           .toEpochMilli();
+
+            long millisecondsBeforeEpoch = 719162 * 24 * 60 * 60 * 1000; // daysBeforeEpoch converted to milliseconds
+            long ticksPerMillisecond = 10_000; // number of ticks per millisecond
+            long ticks = millis * ticksPerMillisecond + millisecondsBeforeEpoch;
```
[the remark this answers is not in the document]
bernardnormier: We should indeed add a Common helper, like in other languages.

java/Ice/bidir/client/src/main/java/com/example/ice/bidir/client/MockAlarmClock.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public MockAlarmClock() {
+        _stopPressed = new CompletableFuture<>();
+    }
+
+    public CompletableFuture<Void> stopPressed() {
+        return _stopPressed;
+    }
+
+    /// <inheritdoc/>
```
bernardnormier, line 27: No a valid Java doc-comment.

java/Ice/bidir/client/src/main/java/com/example/ice/bidir/client/MockAlarmClock.java
```diff
@@ -0,0 +1,41 @@
...
+    // Implements the abstract method Ring from the AlarmClockDisp_ class generated by the Slice compiler.
```
bernardnormier, line 28: lowercase ring, and it's from interface AlarmClock.

java/Ice/bidir/client/src/main/java/com/example/ice/bidir/client/MockAlarmClock.java
```diff
@@ -0,0 +1,41 @@
...
+    public ButtonPressed ring(String message, Current current) {
+        System.out.println("Dispatching ring request { message = '" + message + "' }");
```
bernardnormier, line 30: Should we use `String.format`?

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
bernardnormier, line 20: Incorrect doc-comments.

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
```diff
@@ -0,0 +1,66 @@
...
+        AlarmClockPrx alarmClock = AlarmClockPrx.uncheckedCast(connection.createProxy(new Identity("alarmClock", "")));
+
+        // Start a background task to ring the alarm clock.
+        CompletableFuture.runAsync(() -> {
+            try {
+                // Wait until the specified time.
+                Duration delay = Duration.between(ZonedDateTime.now(ZoneOffset.UTC), timeStampDateTime);
+                if (!delay.isNegative() && !delay.isZero()) {
+                    Thread.sleep(delay.toMillis());
+                }
```
InsertCreativityHere, line 50: Could just do this instead, no need to create the `Duration`:
```suggestion
                if (ZonedDateTime.now(ZoneOffset.UTC).isAfter(timeStampDateTime)) {
                    Thread.sleep(delay.toMillis());
                }
```
Or to be even more technically correct you could do `!isBefore` instead of `isAfter`:
```suggestion
                if (!ZonedDateTime.now(ZoneOffset.UTC).isBefore(timeStampDateTime)) {
                    Thread.sleep(delay.toMillis());
                }
```

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
```diff
@@ -0,0 +1,66 @@
...
+                // First ring. This invocation reuses the connection established by the client.
+                ButtonPressed buttonPressed = alarmClock.ring("It's time to wake up!");
+
+                // Keep ringing every 10 seconds until the user presses the stop button.
+                while (buttonPressed == ButtonPressed.Snooze) {
+                    Thread.sleep(Duration.ofSeconds(10).toMillis());
+                    buttonPressed = alarmClock.ring("No more snoozing!");
+                }
+                System.out.println("Client pressed Stop on alarm clock.");
+            } catch (Exception ex) {
```
bernardnormier, line 61: It's not immediately clear to me why we need this try/catch block. THe C# demo doesn't have one.

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
```diff
@@ -0,0 +1,66 @@
...
+                if (!delay.isNegative() && !delay.isZero()) {
+                    Thread.sleep(delay.toMillis());
+                }
+
+                // First ring. This invocation reuses the connection established by the client.
+                ButtonPressed buttonPressed = alarmClock.ring("It's time to wake up!");
+
+                // Keep ringing every 10 seconds until the user presses the stop button.
+                while (buttonPressed == ButtonPressed.Snooze) {
+                    Thread.sleep(Duration.ofSeconds(10).toMillis());
```
InsertCreativityHere, line 57: I guess this is more readable, but I also think every Java developer knows that `sleep` takes milliseconds by default, and almost every `Thread.sleep` I've seen would just do this instead:
```
Thread.sleep(10_000);
```
But feel free to leave as-is though, it's not like it's wrong or anything

InsertCreativityHere (approved):

java/Ice/bidir/server/src/main/java/com/example/ice/bidir/server/BidirWakeUpService.java
pepone: I will update it to just catch `InterruptedException` thrown from sleep

PR #587, opened 2025-09-18
Add Java callback demo

pepone:
This PR ports Ice/callback demo to Java. It move the time utils to a top-level project included by bidir and callback builds.


InsertCreativityHere (approved):
Looks good to me!

PR #588, opened 2025-09-18
Update Ice/config in all languages

bernardnormier:
This PR updates Ice/config to use initialize(initData), without args.


cpp/Ice/config/Server.cpp
```diff
@@ -16,14 +16,18 @@ main(int argc, char* argv[])
...
+
+    // Create a Properties object from the command line arguments and the config file properties; Ice.* properties and
+    // other reserved properties set in argc/argv override the config file properties.
+    auto properties = Ice::createProperties(argc, argv, configFileProperties);
 
     // Create an Ice communicator. We'll use this communicator to create an object adapter.
-    // The communicator gets its properties from initData.properties; Ice.* properties and other reserved properties
-    // set in argc/argv override these properties.
-    Ice::CommunicatorPtr communicator = Ice::initialize(argc, argv, initData);
+    // The communicator gets its properties from initData.properties.
```
pepone, line 27: Should we say "the properties object." like in C#/Java?

pepone (approved):

PR #589, opened 2025-09-18
Remove pragma once from demo Slice files

bernardnormier:
These Slice files are never included in other Slice files.

As a result of this update, there is no preprocessing directive in any demo Slice file.


bernardnormier:
The most natural "include demo" would be a Glacier2/Session demo. See #590.

PR #592, opened 2025-09-19
Simplify top-level READMEs

bernardnormier:
This PR removes non-essential and occasionally outdated information from the top-level READMEs.

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -3,19 +3,6 @@
 This folder contains example Ice applications written in Ruby. Each example is a simple client application that
 demonstrates a specific feature or programming technique.
 
-> Ice for Ruby is supported on Linux and macOS platforms.
-
-## Building Requirements
-
-The demos in the main branch require the latest Ice for Ruby nightly build. To install it, run:
-
-```shell
-MAKEFLAGS=-j12 gem install zeroc-ice --source https://download.zeroc.com/nexus/repository/rubygems-nightly --user \
-    --pre --version '>= 3.8.0.pre'
-```
-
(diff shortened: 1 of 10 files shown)


externl (approved):

ruby/README.md
```diff
@@ -3,19 +3,6 @@
 This folder contains example Ice applications written in Ruby. Each example is a simple client application that
 demonstrates a specific feature or programming technique.
 
-> Ice for Ruby is supported on Linux and macOS platforms.
```
InsertCreativityHere, line 6: I agree that listing "MATLAB R2024a" is too far, but I think this kind of 'supported platforms' sentence should be moved to `NIGHTLY.md` instead of completely deleted.

It definitely doesn't need to live in this README, but I think it would make sense to have in
[NIGHTLY.md#ice-for-ruby](https://github.com/zeroc-ice/ice/blob/main/NIGHTLY.md#ice-for-ruby)
Otherwise it's not obvious to users that this isn't supported on Windows.

swift/README.md
```diff
@@ -3,14 +3,6 @@
 This folder contains example Ice applications written in Swift. Each example is a simple client-server application that
 demonstrates a specific feature or programming technique.
 
-> Ice for Swift is only supported on macOS and iOS platforms.
```
InsertCreativityHere, line 6: Same here.
All good removing from the README, but I would add this to `NIGHTLY.md` under "Ice for Swift".

InsertCreativityHere (approved):

pepone (approved):

PR #599, opened 2025-09-22
Upgrade Nightly repository configuration

pepone:
This PR updates Gradle maven repository configuration, ZeroC Maven Nightly should be only used for ZeroC artifacts.


externl (approved):

bernardnormier (approved):
I assume we need to make the same update to the instructions in NIGHTLY.md?

PR #608, opened 2025-09-24
Java README minor fixes

pepone:
A minor cleanup for Java READMEs


java/Glacier2/greeter/README.md
```diff
@@ -22,8 +22,6 @@ This project uses Gradle and the [Application plugin]. To build the client and s
 ./gradlew build
 ```
 
-This will compile each project and install the distributions in the `build/install` subdirectory of each project.
```
pepone, line 25: That is a leftover from when we use the `dist` target instead of build.

bernardnormier (approved):

PR #623, opened 2025-09-26
Fix Java Ice/customError build failure

pepone:
Fixes a build failure in Ice/customError related to initialize updates.


java/Ice/customError/client/src/main/java/com/example/ice/customerror/client/Client.java
```diff
@@ -17,10 +18,11 @@ public static void main(String[] args) {
         // Java applications that unmarshal classes or exceptions, when the classes/exceptions or their enclosing
         // module(s) are remapped using java:identifier, like in this demo.
         var initData = new InitializationData();
+        initData.properties = new Properties(args);
```
bernardnormier, line 21: I would move the properties assignment after the sliceLoader assignment, because of the comment above.

bernardnormier (approved):

PR #635, opened 2025-10-01
Rename Swift demos directories to use PascalCase

### LINKED ISSUE #538: Fix names of Swift demo directories
They should all be camel case (Ice/greeter, Ice/forwarder etc.) or all pascal case (Ice/GreeterApp).

externl:
Fixes #538


pepone (commented):
We need to update the links to the folder in the main README, GitHub is case sensitive.

pepone (approved):

PR #645, opened 2025-10-16
Update Python demo servers to use async boilerplate

pepone:


externl (commented):
What about `Ice/greeter/server`?

python/Ice/config/server/main.py
```diff
@@ -1,24 +1,35 @@
...
-def main():
+async def main():
     # Load the contents of the config.server file into a Properties object.
     configFileProperties = Ice.createProperties()
     configFileProperties.load("config.server")
 
+    initData = Ice.InitializationData()
+
     # Create a Properties object from the command line arguments and the config file properties; Ice.* properties and
     # other reserved properties set in sys.argv augment or override the config file properties.
```
bernardnormier: set in the sys.argv command-line arguments (not augment)

python/Ice/greeter/server_asyncio/main.py
```diff
@@ -13,7 +14,7 @@ async def main():
 
     # Create an Ice communicator. We'll use this communicator to create an object adapter. We enable asyncio
     # support by passing the current event loop to initialize.
-    async with Ice.initialize(eventLoop=loop) as communicator:
+    async with Ice.initialize(sys.argv, eventLoop=loop) as communicator:
```
bernardnormier: We should rename this server in a follow-up PR. Since everything uses asyncio, we should not name a server "server_asyncio".

PR #647, opened 2025-10-17
Add Glacier2/session Demo to Java

InsertCreativityHere:
This PR won't work until the slice-tools snapshots pulls in this fix tonight: https://github.com/zeroc-ice/ice/pull/4561

### DIFF
--- a/csharp/Glacier2/Session/Server/IPokeStore.cs
+++ b/csharp/Glacier2/Session/Server/IPokeStore.cs
@@ -6,12 +6,12 @@ namespace Server;
 internal interface IPokeStore
 {
     /// <summary>Saves the Pokémon collection for a specific user.</summary>
-    /// <param name="userId">The user's ID.</param>
+    /// <param name="userId">The user ID.</param>
     /// <param name="pokemon">The Pokémon collection to save.</param>
     void SaveCollection(string userId, IReadOnlyCollection<string> pokemon);
 
     /// <summary>Retrieves the Pokémon collection for a specific user.</summary>
-    /// <param name="userId">The user's ID.</param>
+    /// <param name="userId">The user ID.</param>
     /// <returns>The saved Pokémon collection, or null if no collection was saved for <paramref name="userId"/>.
     /// </returns>
     IReadOnlyCollection<string>? RetrieveCollection(string userId);


csharp/Glacier2/Session/Server/IPokeStore.cs
```diff
@@ -6,12 +6,12 @@ namespace Server;
 internal interface IPokeStore
 {
     /// <summary>Saves the Pokémon collection for a specific user.</summary>
-    /// <param name="userId">The user ID.</param>
+    /// <param name="userId">The user's ID.</param>
```
bernardnormier: I would keep `user ID`. It's a common term, much more common that `user's ID`.

java/Glacier2/session/server/src/main/java/com/example/glacier2/session/server/PokeStore.java
```diff
@@ -0,0 +1,24 @@
...
+    void saveCollection(String userId, List<String> pokemon);
+
+    /**
+     * Retrieves the Pokémon collection of a specific user.
+     *
+     * @param userId the user's ID
+     * @return the saved Pokémon collection, or {@code null} if no collection was saved for {@code userId}.
+     *     The returned list is immutable. Attempting to modify it will throw an UnsupportedOperationException.
+     */
+    List<String> retrieveCollection(String userId);
```
pepone: I guess this a shortcoming of the JDK that doesn't provide an immutable list interface. Here we are documenting the expected behavior of the implementation but we cannot enforce it.

pepone (approved):

java/Glacier2/session/server/src/main/java/com/example/glacier2/session/server/InMemoryPokeStore.java
```diff
@@ -0,0 +1,23 @@
...
+package com.example.glacier2.session.server;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class InMemoryPokeStore implements PokeStore {
+    private final Map<String, List<String>> _store = new HashMap<>();
+
+    public void saveCollection(String userId, List<String> pokemon) {
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

PR #653, opened 2025-10-28
Delete Redundant Checkstyle Action from GitHub CI

InsertCreativityHere:
We're currently running Checkstyle over the repo twice, once in the main `java build`, and another in a stand-alone `checkstyle `action. This PR deletes the standalone checkstyle action. There's no need to run it twice.

----

**Outdated:**

We made this change in the main Ice repository, but I forgot to do the same for the demos.
See https://github.com/zeroc-ice/ice/pull/4359

We have a separate action for running `checkstyle`, so we don't want the main Java build running it as well.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -8,33 +8,6 @@ on:
     branches: ["main"]
 
 jobs:
-  checkstyle:
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
-      - name: Setup Java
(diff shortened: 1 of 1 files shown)


InsertCreativityHere:
> I don't understand why we prefer this setup. To me seems more logical to keep the default and don't run checkStyle as a separate workflow.

I think it makes sense for Ice, where it's clearer to see the builds passed and only checkstyle failed,
as opposed to seeing every build failed.

For the demos, I'd be fine just getting rid of the checkstyle action and letting main CI handle it.
I just don't like that right now we're running it twice is all.

PR #655, opened 2025-10-29
C++ Glacier2/session demo

pepone:


cpp/Glacier2/session/Client.cpp
```diff
@@ -0,0 +1,132 @@
...
+        cout << "\t" << pokemon << endl;
+    }
+
+    if (inventory.size() > 10)
+    {
+        cout << "Oh no! All the Pokémon escaped!" << endl;
+        pokeBox->releaseAll();
+    }
+
+    // Exiting, closing the connection, or calling destroyAsync on the session terminates both PokeSession and the
```
externl:
```suggestion
    // Exiting, closing the connection, or calling `destroy(Async)` on the session terminates both PokeSession and the
```

cpp/Glacier2/session/IPokeStore.h
```diff
@@ -0,0 +1,32 @@
...
+
+        /// Retrieves the Pokémon collection for a specific user.
+        /// @param userId The user ID.
+        /// @returns The saved Pokémon collection, or an empty list if no collection was saved for @p userId.
+        virtual CatchThemAll::PokemonList retrieveCollection(const std::string& userId) const = 0;
+    };
+
+    using IPokeStorePtr = std::shared_ptr<IPokeStore>;
+}
+#endif
```
InsertCreativityHere: Usually an extra line separating the `endif` from the actual code.
```suggestion

#endif
```

cpp/Glacier2/session/InMemoryPokeStore.cpp
```diff
@@ -0,0 +1,18 @@
+
+#include "InMemoryPokeStore.h"
+
+using namespace Server;
+using namespace std;
+
+void
+InMemoryPokeStore::saveCollection(const string& userId, CatchThemAll::PokemonList pokemon)
+{
+    _store[userId] = std::move(pokemon);
```
InsertCreativityHere: This implementation seems different than the other languages.
In the others passing an empty list causes us to remove the entry, whereas here we never remove entries.

I don't think this is 'wrong', just different from the others.

cpp/Glacier2/session/SharedPokeBox.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) ZeroC, Inc.
+
+#include "IPokeStore.h"
+#include "IUserIdResolver.h"
+#include "PokeBox.h"
+
+#include <string>
+#include <vector>
```
InsertCreativityHere: Don't see where we need `vector` here either:
```suggestion
```

PR #657, opened 2025-10-30
Update DataStorm sample filter demo

pepone:
This PR updates the DataStorm/sampleFilter demo to show how to create a custom sample filter.

### DIFF
--- a/cpp/DataStorm/sampleFilter/Writer.cpp
+++ b/cpp/DataStorm/sampleFilter/Writer.cpp
@@ -46,7 +46,7 @@ main(int argc, char* argv[])
     auto shutdownPromise = std::promise<void>();
     auto shutdownFuture = shutdownPromise.get_future();
     ctrlCHandler.setCallback(
-        [&ctrlCHandler, &shutdownPromise](int)
+        [&shutdownPromise](int)
         {
             std::cout << "Shutting down..." << std::endl;
             shutdownPromise.set_value();


cpp/DataStorm/sampleFilter/Reader.cpp
```diff
@@ -4,40 +4,57 @@
...
-        cerr << ex.what() << endl;
-        return 1;
+        printSample(sample);
     }
+}
+
+int
+main(int argc, char* argv[])
+{
+    // CtrlCHandler must be created before the communicator is created or any other threads are started.
```
InsertCreativityHere: Same here:
```suggestion
    // CtrlCHandler is a helper class that handles Ctrl+C and similar signals. It must be constructed at the beginning
    // of the program, before creating an Ice communicator or starting any thread.
```

cpp/DataStorm/sampleFilter/Reader.cpp
```diff
@@ -4,40 +4,57 @@
...
+    // Instantiates DataStorm node.
+    DataStorm::Node node{argc, argv};
+
+    ctrlCHandler.setCallback(
+        [&ctrlCHandler, &node](int)
+        {
+            std::cout << "Shutting down..." << std::endl;
+            node.shutdown();
+            // Reset the callback to nullptr to avoid calling it again.
+            ctrlCHandler.setCallback(nullptr);
```
InsertCreativityHere: Is this necessary? I haven't seen it in other demos, and `setCallback` locks a non-reentrant mutex internally.
Locking mutexes in a signal handler always feels a little dicey to me.

cpp/DataStorm/sampleFilter/Writer.cpp
```diff
@@ -3,44 +3,85 @@
...
-        // Configure writers to not clear the history. We want the readers to receive all the writer samples.
-        topic.setWriterDefaultConfig({std::nullopt, std::nullopt, DataStorm::ClearHistoryPolicy::Never});
+    // Instantiates the "temperature" topic. The topic uses strings for keys and floats for values.
+    DataStorm::Topic<string, float> topic{node, "temperature"};
 
-        // Instantiate the foo writer and wait for a reader to connect.
-        auto writer = DataStorm::makeSingleKeyWriter(topic, "foo");
-        topic.waitForReaders();
+    // Create a sample filter that only reports temperatures outside of the given temperature range.
+    // The filter criteria TemperatureRange is defined in TemperatureRange.ice Slice file.
```
InsertCreativityHere: For consistency if you do the other one:
```suggestion
    // The filter criteria `TemperatureRange` is generated from the TemperatureRange.ice Slice file.
```

cpp/DataStorm/sampleFilter/Writer.cpp
```diff
@@ -3,44 +3,85 @@
...
+    // Shutdown the writer on Ctrl-C.
+    auto shutdownPromise = std::promise<void>();
+    auto shutdownFuture = shutdownPromise.get_future();
+    ctrlCHandler.setCallback(
+        [&ctrlCHandler, &shutdownPromise](int)
+        {
+            std::cout << "Shutting down..." << std::endl;
+            shutdownPromise.set_value();
+            // Reset the callback to nullptr to avoid calling it again.
+            ctrlCHandler.setCallback(nullptr);
```
InsertCreativityHere: I'm also dubious of the `setCallback` mutex locking here.

Here, it makes more sense, since `set_value` isn't allowed to be called multiple times.
But if the user is spamming Ctrl-C over and over, I think it's fine to die with an exception.

With the current logic, it sure looks to me like spamming Ctrl-C could cause a dead-lock in `setCallback`
(if a 2nd interrupt is triggered while the handler is still responding to a 1st signal).

InsertCreativityHere (approved):
Looks good to me.

cpp/DataStorm/sampleFilter/Reader.cpp
pepone: Copied from one of the demos, but seems we don't have this in most of the demos.

pepone: Removed

cpp/DataStorm/sampleFilter/Reader.cpp
```diff
@@ -4,40 +4,56 @@
...
 
-        // Instantiates the "hello" topic. The topic uses strings for keys and values.
-        DataStorm::Topic<string, string> topic{node, "hello"};
+    // Instantiates DataStorm node.
+    DataStorm::Node node{argc, argv};
 
-        // Configure readers to never clear the history. We want to receive all the samples written by the writers.
-        topic.setReaderDefaultConfig({std::nullopt, std::nullopt, DataStorm::ClearHistoryPolicy::Never});
+    ctrlCHandler.setCallback(
+        [&ctrlCHandler, &node](int)
```
InsertCreativityHere: Was reviewing your other demo and realized that my suggestion was incomplete.
You can also remove the capture of `ctrlCHandler` now, since it isn't being set to `nullptr` anymore:
```suggestion
        [&node](int)
```

PR #659, opened 2025-10-30
DataStorm partial update demo

pepone:
This PR adds a new DataStorm demo that showcase partial update feature.

The DataStorm/stock demo already shows partial updates the idea is to replace DataStorm/stock with this simpler demo.


cpp/DataStorm/partialUpdate/Writer.cpp
```diff
@@ -0,0 +1,134 @@
...
+            shutdownPromise.set_value();
+            // Reset the callback to nullptr to avoid calling it again.
+            ctrlCHandler.setCallback(nullptr);
+        });
+
+    // Initialize random number generators.
+    std::mt19937 gen{std::random_device{}()};
+    std::uniform_int_distribution<> atmosphericDist{0, 10};
+
+    ClearSky::AtmosphericConditions conditions{21.5f, 50.0f};
```
externl: We should comment these two lines as thy interact with datastorm. Are these some kind of initial conditions? This is a full update?

cpp/DataStorm/partialUpdate/README.md
```diff
@@ -1,10 +1,11 @@
-# DataStorm Stock
+# DataStorm partialUpdate
 
-This demo illustrates the use of partial updates. The writer adds stocks to the
-topic and sends partial updates to update the stock price or volume. The reader
-prints out the stock information and partial updates.
+This demo illustrates the use of **partial updates** in DataStorm.
```
bernardnormier: I would remove the `**`, we don't use this anywhere else.

cpp/DataStorm/partialUpdate/AtmosphericConditions.ice
```diff
@@ -0,0 +1,21 @@
...
+module ClearSky
+{
+
+    enum UpdateTag
+    {
+        TemperatureUpdated,
+        HumidityUpdated
+    }
+
+    /// Represents the atmospheric conditions measured by a sensor.
```
bernardnormier: Unless there is compelling reason to use a struct, I would use a class like all the other AtmosphericConditions.

cpp/DataStorm/partialUpdate/Writer.cpp
```diff
@@ -0,0 +1,134 @@
...
+    float minTemp = 19.0f;
+    float maxHumidity = 55.0f;
+    float minHumidity = 45.0f;
+
+    bool humidityIncreasing = true;
+    bool tempIncreasing = true;
+
+    while (true)
+    {
+        // Generate random changes where either temperature or humidity changes about 50% of the time.
```
bernardnormier: That's too complicated and distracts from the focus of this demo - DataStorm.

It also gives the wrong impression that "partial update" is about diffs, which is NOT true.

PR #660, opened 2025-10-31
Rework custom encoding demo

pepone:
This PR reworks the DataStorm demo for custom encoding, previously named clock, renamed to customEncoding.

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -7,7 +7,7 @@ demonstrates a specific feature or programming technique.
 |----------------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
 | [DataStorm Custom Encoding](./DataStorm/customEncoding/) | Shows how to implement a custom encoder and decoder.                                                        |
 | [DataStorm KeyFilter](./DataStorm/keyFilter/)            | Shows how to use a key filters.                                                                             |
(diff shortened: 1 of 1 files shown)


cpp/DataStorm/customEncoding/TimePoint.h
```diff
@@ -0,0 +1,50 @@
+#ifndef TIME_POINT_H
+#define TIME_POINT_H
+
+#include <DataStorm/DataStorm.h>
+#include <Ice/Ice.h>
+
+namespace DataStorm
```
pepone: Moved the Encoded/Decoder template from Reader/Writer to this file.

I also documented them and updated to use bitwise operators instead of division which I find much clear for a binary encoding.

cpp/DataStorm/customEncoding/Reader.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <chrono>
+#include <iostream>
+
+#include "TimePoint.h"
+
+using namespace std;
+
+int
+main(int argc, char* argv[])
+{
```
pepone: Removed exception handling from the reader and writer, as it’s unnecessary for these simple demos and consistent with our other examples.

InsertCreativityHere (approved):
Looks good to me!

externl (approved):

PR #663, opened 2025-11-12
Various VSCode fixes

externl:
- Load Pokemon cspell dictionary from CDN as installing with NPM is a pain
- Fix list of cmake source dirs, and add an action to test.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -86,8 +86,8 @@ jobs:
 
           # For each directory with a CMakeLists.txt file, ensure it is listed in cmake.sourceDirectory
           find . -name CMakeLists.txt -type f | while read -r file; do
-              dir=$(dirname "$file" | sed 's#^\./##')
-              if ! jq -r '.["cmake.sourceDirectory"][]' .vscode/settings.json | sed 's#\${workspaceFolder}/##' | grep -qx "$dir"; then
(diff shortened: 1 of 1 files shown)


.github/workflows/cpp.yml
```diff
@@ -61,3 +61,34 @@ jobs:
...
+              if [ ! -f "$dir/CMakeLists.txt" ]; then
+                  echo "Directory '$dir' listed in .vscode/settings.json does not exist or is missing CMakeLists.txt"
+                  exit 1
+              fi
+          done
+
+          # For each directory with a CMakeLists.txt file, ensure it is listed in cmake.sourceDirectory
+          find . -name CMakeLists.txt -type f | while read -r file; do
+              dir=$(dirname "$file" | sed 's#^\./##')
+              if ! jq -r '.["cmake.sourceDirectory"][]' .vscode/settings.json | sed 's#\${workspaceFolder}/##' | grep -qx "$dir"; then
```
InsertCreativityHere: Really doesn't matter,
but why not just replace `.` with `${workspaceFolder}` in one shot, instead of using 2 `sed`s:

```suggestion
              dir=$(dirname "$file" | sed 's#^\./#\${workspaceFolder}/#')
              if ! jq -r '.["cmake.sourceDirectory"][]' .vscode/settings.json | grep -qx "$dir"; then
```

Issue #664, opened 2025-11-16
Update Python demos to use uv


pepone:
Using [uv](https://docs.astral.sh/uv/) greatly simplifies the setup for the demos.

If project have a pyproject.toml like:

```
[project]
name = "client"
version = "0.1.0"
description = "Ice Greeter client"
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["zeroc-ice"]


[tool.uv]
# Allow pre-releases globally (so the nightly zeroc-ice can be selected)
prerelease = "allow"

# Configure the ZeroC nightly index as an additional index,
# searched *before* the default PyPI index.
[[tool.uv.index]]
name = "zeroc-nightly"
url = "https://download.zeroc.com/nexus/repository/pypi-nightly/simple/"
```

Building an running the demo:

The virtual environment is automatically create by **uv**, running a command, using uv run ensures the required dependencies are installed.

```
jose@m1 server % uv run slice2py ../slice/Greeter.ice 
Using CPython 3.13.9 interpreter at: /opt/homebrew/opt/python@3.13/bin/python3.13
Creating virtual environment at: .venv
Installed 1 package in 7ms
```

Running the server:

```
jose@m1 server % uv run main.py 
Listening on port 4061...
```

PR #665, opened 2025-11-17
Uv python

pepone:
This PR upgrades Ice/greeter python demo to use [uv](https://docs.astral.sh/uv/) for building and running the demo. See #664

If accepted I will upgrade other Python demos too.

### DIFF
--- a/python/Ice/greeter/client/pyproject.toml
+++ b/python/Ice/greeter/client/pyproject.toml
@@ -2,7 +2,6 @@
 name = "client"
 version = "0.1.0"
 description = "Ice Greeter Client"
-readme = "README.md"
 requires-python = ">=3.12"
 dependencies = ["zeroc-ice"]
 
--- a/python/Ice/greeter/server_asyncio/pyproject.toml
+++ b/python/Ice/greeter/server_asyncio/pyproject.toml
@@ -2,7 +2,6 @@
 name = "server-asyncio"
 version = "0.1.0"
 description = "Ice Greeter Asynchronous Server using asyncio"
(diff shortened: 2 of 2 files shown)


python/Ice/greeter/client/.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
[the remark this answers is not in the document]
pepone: I added 3.13 to ensure uv runs with the python version which is the one we are building wheel packages for. If we set it to 3.12 it will trigger an source build.

I create https://github.com/zeroc-ice/ice/pull/4651 to build additional wheel packages.

python/Ice/greeter/client/README.md
InsertCreativityHere: This file is empty, probably snuck into a commit and should be deleted.

pepone:
> Not familiar with uv, but the changes look good to me.

Main thing to review here is whether or not moving the demos to use the new tool is ok. I updated greeter as an example of what it will look like using uv.

python/Ice/greeter/client/README.md
InsertCreativityHere: This file is empty, probably snuck into a commit and should be deleted.

InsertCreativityHere:
> Main thing to review here is whether or not moving the demos to use the new tool is ok.

Yeah, I'm just not familiar enough with the Python ecosystem, but seems like a newer version of `poetry` to me.
But I did clone and build with `uv` locally (after commenting out the signal code...), and it's definitely faster and simpler than our current setup, so fine with me!

PR #669, opened 2025-11-19
Use uv for running and building Ice Python demos

pepone:
Updates all Python demos to use uv

### DIFF
--- a/python/Glacier2/callback/README.md
+++ b/python/Glacier2/callback/README.md
@@ -26,13 +26,13 @@ Navigate to the `server` directory.
 
 Use the Slice-to-Python compiler to generate Python code from the `AlarmClock.ice` file:
 
-```bash
+```shell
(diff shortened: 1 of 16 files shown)


python/Glacier2/callback/README.md
```diff
@@ -13,54 +13,30 @@ flowchart LR
     g --connection1--> s[Server:4061<br>hosts WakeUpService] --connection2--> g
 ```
 
-We recommend running each program in a separate Python virtual environment.
-If you are new to Python virtual environments, see [Python Virtual Environments].
-
-## Ice prerequisites
+## Prerequisites
 
+- [uv] Python package and project manager.
```
bernardnormier: This should be:
```
- Install uv. See [Installing uv].
```

Where Installing uv = https://docs.astral.sh/uv/getting-started/installation/

PR #677, opened 2025-12-10
Move signal.signal before await shutdownCompleted

bernardnormier:
Just like in the C# demos: it's clearly to perform the signal setup just before we wait until Ctrl+C.

In some languages (such as C++ and Swift), we need to do it at the beginning of the program. But not in Python/C#.


python/Glacier2/callback/server/main.py
```diff
@@ -31,6 +28,9 @@ async def main():
         adapter.activate()
         print("Listening on port 4061...")
 
+        # Shutdown the communicator when the user presses Ctrl+C.
+        signal.signal(signal.SIGINT, lambda _signum, _frame: communicator.shutdown())
```
bernardnormier: The `_` means not used on purpose according to Claude.

PR #679, opened 2025-12-11
Remove signal from greeter demo (Python)

bernardnormier:
I find this style nicer. If approved, I'll update all Python demos in a follow-up PR.


externl (approved):

pepone:
I can't find any docs on what asynio does on Ctrl-C, specially does it allow  `async with Ice.Communicator(` to complete normally?

pepone (approved):

PR #680, opened 2025-12-11
Replace signal by try/except in Python demos

bernardnormier:


python/IceGrid/greeter/server/main.py
```diff
@@ -32,10 +31,7 @@ async def main():
         adapter.activate()
         print(f"{greeterName} is listening...")
 
-        # Shutdown the communicator when the user presses Ctrl+C.
-        signal.signal(signal.SIGINT, lambda _signum, _frame: communicator.shutdown())
-
-        # Wait until the communicator is shut down.
+        # Wait until the communicator is shut down by IceGrid.
```
bernardnormier: We don't need to handle Ctrl+C in this demo.

bernardnormier: Updated comment to use same comment as in other languages.

externl (approved):

PR #689, opened 2025-12-17
Upgrade C# solutions to SLNX format

### LINKED ISSUE #688: Upgrade C# solutions to slnx format
We should use the new slnx format unless there is a compelling reason to keep the sln format.

pepone:
Fix #688


bernardnormier (approved):
Looks good!

PR #690, opened 2025-12-17
Enable and update CI to run on 3.8 branch

pepone:


pepone:
> what about the cmake config for windows?

I think is already updated in the 3.8 branch. The C++ build worked, is just Java failing because the missing Gradle plugin.

externl (approved):

PR #691, opened 2025-12-17
Add exception handling to C++ Ice/greeter

bernardnormier:
This PR adds a custom terminate handler to the C++ Ice/greeter demo.

The main motivation for this update the default terminate handler on Windows, which calls about without providing any information. With this PR, we get instead:

client
```
C:\Users\bernard\source\repos\ice-demos\cpp\Ice\greeter>build\client
D:\a\ice\ice\cpp\src\Ice\Network.cpp:1865 Ice::ConnectionRefusedException connection refused
stack trace:
  0 00007ffc7761d96a
  1 00007ffc775c4c33
  2 00007ffc775c4b83
  3 00007ffc775bb653
  4 00007ffc77624b70
  5 00007ffc77a6ff7a
  6 00007ffc77b2cdff
  7 00007ffc77b3ab3a
  8 00007ffc7760e8fa
  9 00007ffc7783cebb
 10 00007ffc778374f7
 11 00007ffc77b3f12a
 12 00007ffc77b3ca47
 13 00007ffc77b4818b
 14 00007ffc77b469b7
 15 00007ffcbff82ec5
 16 00007ffcfbcd7374
 17 00007ffcfc37cc91

!! 12/17/25 13:08:50.612 error: communicator not destroyed during global destruction.
```

server
```
C:\Users\bernard\source\repos\ice-demos\cpp\Ice\greeter>build\serveramd
D:\a\ice\ice\cpp\src\Ice\Network.cpp:1332 Ice::SocketException socket error: WSAEADDRINUSE
stack trace:
  0 00007ffc7761d96a
  1 00007ffc77624025
  2 00007ffc77624511
  3 00007ffc775c4ad3
  4 00007ffc77a6e9b8
  5 00007ffc77b2f6f0
  6 00007ffc7780a028
  7 00007ffc77808d97
  8 00007ffc77a934c7
  9 00007ffc77a852c4
 10 00007ffc7759aff5
 11 main at C:\Users\bernard\source\repos\ice-demos\cpp\Ice\greeter\ServerAMD.cpp:28
 12 invoke_main at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:79
 13 __scrt_common_main_seh at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
 14 __scrt_common_main at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:331
 15 mainCRTStartup at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_main.cpp:17
 16 00007ffcfbcd7374
 17 00007ffcfc37cc91

!! 12/17/25 13:11:42.841 error: communicator not destroyed during global destruction.
```

and no abort dialog.

The advantage of this custom terminate handler are:
 - only one extra line of code per C++ program (+ include for common/Terminate.h)
 - no change to flow / horizontal whitespace of the main function
 - factors exception handling to common shared file where we can add comments and do more stuff

Disadvantage:
 - harder to find the exception handling if you're interested in it
 - more advanced C++ code

An alternative would be to enclose the code of each main function in a big try/catch block:
```cpp
try
{
   ... existing code
}
catch (const Ice::Exception& ex)
{
    // We can stream Ice exceptions with operator<<.
    cerr << ex << endl;
    return 1;
}
catch (const std::exception& ex)
{
    cerr << "std::exception: " << ex.what() << endl;
    return 1;
}
```

The advantage of this style are:
 - simple C++ code, easy to find and follow

Disadvantage:
 - many extra lines of exception handling code in each demo
 - horizontal spacing of the demo code

Overall, I lean toward the custom terminate handler provided by this PR.

### DIFF
--- a/cpp/common/Terminate.h
+++ b/cpp/common/Terminate.h
@@ -10,8 +10,8 @@
 
 namespace Terminate
 {
-    /// Custom terminate handler that prints unhandled exceptions to std::cerr.
-    /// This custom handler is primarily useful on Windows where the default terminate handler aborts the application
+    /// Custom terminate handler that prints unhandled exceptions to std::cerr before aborting the program.
+    /// This custom handler is primarily useful on Windows where the default terminate handler aborts the program
     /// without providing any information.
     inline void printCurrentException()
     {
@@ -40,10 +40,7 @@ namespace Terminate
         {
             std::cerr << "Caught unknown exception" << std::endl;
         }
-
-        // std::exit does not call destructors of stack objects such as CommunicatorHolder.
-        // If the communicator is not destroyed at exit, we get a warning message on the console.
-        std::exit(EXIT_FAILURE);
+        std::abort();
     }
 }
 


bernardnormier:
For reference, that's the error dialog we currently get on Windows (prior to this PR).

image

pepone:
Why not add a common main we include in all projects and have each define run.

```
// common/main.cpp
int run(int argc, char* argv[]);

int
main(int argc, char* argv[])
{
    try {
        return run(argc, argv);
    }
    catch (const std::exception& ex)
    {
        cerr << ex.what() << "\n";
        return -1;
    }
}
```

The default control handler makes easy to jump into the debugger on Windows. Replacing it in code that is intended for users to copy doesn't seems like the best default.

bernardnormier:
> I'd rather avoid have a helper. Just do the try/catch.

I prefer to avoid all this verbose exception handling, just like we do in other languages.

PR #754, opened 2026-06-03
Update Ice version to 3.8.2

bernardnormier:


externl (approved):
Looks good. No C#? Also look like we don't have PDBs uploaded yet.

PR #758, opened 2026-06-05
Update 3.8 ci.yml workflow

externl:


pepone (approved):
Looks good, we should add Ubuntu 26.04

PR #784, opened 2026-06-30
Bump openrewrite plugin to 7.35.0, keeping 4-space continuation indent

externl:
Bumps the openrewrite gradle plugin from 7.7.0 to 7.35.0 (supersedes #781).

Newer openrewrite derives its code style from `checkstyle.xml` and lets it win over styles declared in `rewrite.yml`. Our config had no `Indentation` module, so openrewrite fell back to its 8-space continuation indent and wanted to reformat every Java file.

Per review, `checkstyle.xml` is now the single source of truth for the formatting spec:

- Add the full `Indentation` settings and `OperatorWrap` (nl) to `checkstyle.xml`
- Delete the ZeroC style from `rewrite.yml` and its `activeStyle()` activation

Code updated for the newly-enforced checks:

- Rewrap the multi-line `catch` clauses in the secure client/server demos (Checkstyle `Indentation`)
- Replace the empty `if` block in the middleware client with a guard clause (openrewrite `EmptyBlock`)


java/config/checkstyle/checkstyle.xml
```diff
@@ -84,6 +84,13 @@
   <!-- 'TreeWalker' is what Checkstyle uses to traverse the source code's AST. -->
   <!-- Any checks that require some language awareness must be in this module. -->
   <module name="TreeWalker">
+    <!-- Keep in sync with the TabsAndIndentsStyle in ../rewrite.yml. In addition to being enforced
+         by Checkstyle, OpenRewrite derives its indentation style from this module; without it,
+         OpenRewrite falls back to an 8-space continuation indent and reformats the code. -->
+    <module name="Indentation">
+      <property name="basicOffset" value="4"/>
+      <property name="lineWrappingIndentation" value="4"/>
+    </module>
```
InsertCreativityHere: With this current setup, we duplicate our indentation settings between here (`checkstyle.xml`) and rewrite (`rewrite.yml`). Reading through the rewrite docs, it seems the recommended pattern nowadays is to let your Checkstyle be the single source of truth. So, I think we should:
A) Just delete our '"ZeroC Java Style"' defined in `rewrite.yml`
B) Add all the necessary entries to the `indentation` module of our checkstyle file here.
So, this:
```suggestion
    <module name="Indentation">
      <property name="basicOffset" value="4"/>
      <property name="lineWrappingIndentation" value="4"/>
      <property name="caseIndent" value="4"/>
      <property name="throwsIndent" value="4"/>
      <property name="arrayInitIndent" value="4"/>
      <property name="braceAdjustment" value="0"/>
    </module>
```

InsertCreativityHere (commented):
I think the idea is right, but doesn't go far enough in switching from `rewrite.yml` to `checkstyle.xml` as the source-of-truth for our formatting spec.

java/config/checkstyle/checkstyle.xml
externl: Done in 9273497 — the ZeroC style is gone and `checkstyle.xml` now carries the full `Indentation` settings.

I also added an `OperatorWrap` module (`option: nl`). The deleted style carried `OperatorWrapStyle: NL`, and without a matching Checkstyle module OpenRewrite falls back to its own operator-wrap default.

Verified locally: OpenRewrite 7.35.0 wants no reformatting, and checkstyle 10.21.4 passes on all Java files with the new checks enforced.

java/config/checkstyle/checkstyle.xml
```diff
@@ -84,6 +84,21 @@
...
+      <property name="basicOffset" value="4"/>
+      <property name="lineWrappingIndentation" value="4"/>
+      <property name="caseIndent" value="4"/>
+      <property name="throwsIndent" value="4"/>
+      <property name="arrayInitIndent" value="4"/>
+      <property name="braceAdjustment" value="0"/>
+    </module>
+    <module name="OperatorWrap">
+      <property name="option" value="nl"/>
+    </module>
```
InsertCreativityHere: We already have an `OperationWrap` module defined later in this file:
```
    <module name="OperatorWrap">
      <property name="option" value="NL"/>
      <property name="tokens"
               value="BAND, BOR, BSR, BXOR, DIV, EQUAL, GE, GT, LAND, LE, LITERAL_INSTANCEOF, LOR,
                    LT, MINUS, MOD, NOT_EQUAL, PLUS, QUESTION, SL, SR, STAR, METHOD_REF,
                    TYPE_EXTENSION_AND "/>
```
that looks identical to this one, but includes an explicit list of tokens.

A) What are the benefits of having 2 separator `OperatorWrap` modules?
B) What are the differences between the 2 modules, given that one uses a default `tokens` list, and the other explicitly states some of them. Which tokens are covered by one but not another?

externl: Good catch — no benefit, I'd missed the existing module. Removed the duplicate in a613e5f; the existing one alone now drives both Checkstyle and the style OpenRewrite derives.

On B: both use `option: nl`. The default token set is the explicit list minus `METHOD_REF` plus `COLON` — so the duplicate additionally enforced wrapping on ternary colons, while the existing list deliberately skips `COLON` and adds `METHOD_REF`. Re-verified: OpenRewrite wants no reformatting and checkstyle passes on all Java files.

InsertCreativityHere: kewl

java/config/checkstyle/checkstyle.xml
```diff
@@ -84,6 +84,18 @@
   <!-- 'TreeWalker' is what Checkstyle uses to traverse the source code's AST. -->
   <!-- Any checks that require some language awareness must be in this module. -->
   <module name="TreeWalker">
+    <!-- This file is the source of truth for our formatting spec: in addition to being enforced
+         by Checkstyle, OpenRewrite derives its code style from these modules (this one and the
+         'OperatorWrap' module below). Without an 'Indentation' module, OpenRewrite falls back to
+         an 8-space continuation indent and reformats the code. -->
```
InsertCreativityHere: I don't think this comment is necessary here. This is just a regular checkstyle module, nothing interesting. If you want to add a comment about rewrite stuff, maybe remove this comment and add a very short comment to the `rewrite.yml`` file stating that it derives its style from our checkstyle styling. But feel free to use different words, as long as it's short.
```suggestion
```

externl: Done in d0824df — comment removed, one-line note added to `rewrite.yml`.

InsertCreativityHere (approved):
Looks good to me!

PR #798, opened 2026-08-04
Add least-privilege permissions to workflows

externl:
Workflows had no `permissions:` block, so each run inherited the repository default of write on nearly everything (contents, packages, actions, security-events).

- Declare top-level `permissions: contents: read` in all nine workflows; none of them write to the repo, publish packages, or use secrets.
- Pass `--ignore-scripts` to the `markdownlint-cli` and `pyright` global installs, so a compromised dependency cannot run an install hook in CI. Neither tree contains install scripts today.


pepone (approved):
Looks good!

PR #821, opened 2026-08-05
Fix 32-bit overflow and wrong-unit epoch offset in Java Time helper

### LINKED ISSUE #802: [Audit-High] Java shared Time helper: 32-bit overflow and wrong-unit epoch offset break cross-language bidir/callback demos
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), during independent verification of the [2026-08-04 ice-demos correctness audit][audit]; recorded as **NEW-H01** in the [verification report][verif]. Code is pinned to `main` at [`99afbe3`][commit]. **Not yet human-triaged** — please verify before acting.

## Summary

The shared Java `Time` helper — used by the `Ice/bidir`, `Ice/callback`, and `Glacier2/callback` clients and servers — has two independent bugs that violate the Slice timestamp contract ("number of ticks (100 nanoseconds) since January 1, 0001 00:00:00 UTC"):

https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/util/src/main/java/com/example/util/Time.java#L5-L31

1. **The epoch-offset constant overflows 32-bit arithmetic** (line 7):

   ```java
   private static final long MILLIS_BEFORE_EPOCH = 719162 * 24 * 60 * 60 * 1000;
   ```

   Every operand is an `int` literal, so the product wraps before the widening assignment: the intended 62,135,596,800,000 becomes **304,928,768** (~3.5 days).

2. **The offset is applied in the wrong unit** (line 19): a milliseconds offset is added *after* the tick conversion. The correct expression is `(toEpochMilli() + MILLIS_BEFORE_EPOCH) * TICKS_PER_MILLISECOND` — exactly what the JS sibling does with overflow-proof BigInt literals ([`js/Ice/bidir/timeUtil.ts#L3-L9`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/js/Ice/bidir/timeUtil.ts#L1-L10)) and what `cpp/common/Time.h` does via `chrono`.

`toZonedDateTime` (line 29) makes the mirror-image mistakes, so **Java↔Java pairings round-trip and hide both bugs**. Any cross-language pairing is broken:

- **Java client → C++/C#/Swift server:** the timestamp decodes to a date near year 1; `sleep_until` fires immediately, so the alarm rings right away instead of after the 5-second delay.
- **JS client → Java server:** JS sends correct ticks; Java decodes them to a date ~1,969 years in the future, so the alarm never rings. This is a documented normal path — `js/Ice/bidir` is client-only and its README directs users to start a server in a server-capable language such as Java.

## Fix direction

Use a `long` literal (`719162L * 24 * 60 * 60 * 1000`, or compute from `LocalDate.of(1, 1, 1)`), and fold the offset in before the tick multiplication in both conversion directions.

bernardnormier:
Fixes #802.

The shared Java `Time` helper (used by the `Ice/bidir`, `Ice/callback`, and `Glacier2/callback` clients and servers) had two bugs that broke the Slice timestamp contract for any cross-language pairing:

- `MILLIS_BEFORE_EPOCH` was computed with all-`int` arithmetic, wrapping to 304,928,768 ms (~3.5 days) instead of the intended 62,135,596,800,000 ms.
- The offset was applied in milliseconds *after* the tick multiplication (and subtracted before the division in the reverse direction), i.e. in the wrong unit.

Both conversion directions now match the JS (`js/Ice/bidir/timeUtil.ts`) and C++ (`cpp/common/Time.h`) helpers: `(epochMillis + MILLIS_BEFORE_EPOCH) * TICKS_PER_MILLISECOND`.

Verified by compiling the fixed helper and checking that:

- the Unix epoch converts to exactly 621,355,968,000,000,000 ticks (the well-known .NET `DateTime.UnixEpoch.Ticks` value) and back;
- a current timestamp round-trips exactly;
- `java/Ice/bidir` builds.

### DIFF
--- a/java/util/src/main/java/com/example/util/Time.java
+++ b/java/util/src/main/java/com/example/util/Time.java
@@ -26,7 +26,7 @@ public class Time {
      * @return The zoned date time.
      */
     public static java.time.ZonedDateTime toZonedDateTime(long timeStamp) {
-        long epochMilli = timeStamp / TICKS_PER_MILLISECOND - MILLIS_BEFORE_EPOCH;
+        long epochMilli = (timeStamp / TICKS_PER_MILLISECOND) - MILLIS_BEFORE_EPOCH;
         return java.time.Instant.ofEpochMilli(epochMilli).atZone(java.time.ZoneOffset.UTC);
     }
 
--- a/java/util/src/main/java/com/example/util/Time.java
+++ b/java/util/src/main/java/com/example/util/Time.java
@@ -3,10 +3,10 @@
 package com.example.util;
 
 public class Time {
-    // daysBeforeEpoch converted to milliseconds
-    private static final long MILLIS_BEFORE_EPOCH = 719162 * 24 * 60 * 60 * 1000;
+    // Days before epoch converted to milliseconds.
+    private static final long MILLIS_BEFORE_EPOCH = 719_162L * 24 * 60 * 60 * 1000;
 
-    // number of ticks per millisecond
+    // Number of ticks per millisecond.
     private static final long TICKS_PER_MILLISECOND = 10_000;
 
     /**
@@ -16,7 +16,7 @@ public class Time {
      * @return The time stamp.
      */
     public static long toTimeStamp(java.time.ZonedDateTime dateTime) {
-        return dateTime.toInstant().toEpochMilli() * TICKS_PER_MILLISECOND + MILLIS_BEFORE_EPOCH;
+        return (dateTime.toInstant().toEpochMilli() + MILLIS_BEFORE_EPOCH) * TICKS_PER_MILLISECOND;
     }
 
     /**
@@ -26,7 +26,7 @@ public class Time {
      * @return The zoned date time.
      */
     public static java.time.ZonedDateTime toZonedDateTime(long timeStamp) {
-        long epochMilli = (timeStamp - MILLIS_BEFORE_EPOCH) / TICKS_PER_MILLISECOND;
+        long epochMilli = timeStamp / TICKS_PER_MILLISECOND - MILLIS_BEFORE_EPOCH;
         return java.time.Instant.ofEpochMilli(epochMilli).atZone(java.time.ZoneOffset.UTC);
     }
 


java/util/src/main/java/com/example/util/Time.java
```diff
@@ -3,10 +3,10 @@
 package com.example.util;
 
 public class Time {
-    // daysBeforeEpoch converted to milliseconds
-    private static final long MILLIS_BEFORE_EPOCH = 719162 * 24 * 60 * 60 * 1000;
+    // Days before epoch converted to milliseconds.
```
InsertCreativityHere, line 6: Why not just say:
```suggestion
    // The number of milliseconds before the epoch.
```

The old comment only mentioned days because it's the name of a variable in our C++ code:
```
    inline std::int64_t toTimeStamp(const std::chrono::system_clock::time_point& timePoint)
    {
        const int daysBeforeEpoch = 719162;
```
But if we aren't talking about this variable, then there's no reason to mention anything about days.

java/util/src/main/java/com/example/util/Time.java
```diff
@@ -26,7 +26,7 @@ public static long toTimeStamp(java.time.ZonedDateTime dateTime) {
      * @return The zoned date time.
      */
     public static java.time.ZonedDateTime toZonedDateTime(long timeStamp) {
-        long epochMilli = (timeStamp - MILLIS_BEFORE_EPOCH) / TICKS_PER_MILLISECOND;
+        long epochMilli = timeStamp / TICKS_PER_MILLISECOND - MILLIS_BEFORE_EPOCH;
```
InsertCreativityHere: This is correct, Java will ensure that the operations run in the correct order,
But it would be more readable to keep parenthesis around the division.
```suggestion
        long epochMilli = (timeStamp / TICKS_PER_MILLISECOND) - MILLIS_BEFORE_EPOCH;
```
Especially since this is for the demos, where readability is important.

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):
> **AI-generated review** — by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), verifying the arithmetic and cross-checking against the JS and C++ reference helpers.

### Correctness — verified

- **The constant is now computed in 64-bit arithmetic**: `719_162L * 24 * 60 * 60 * 1000` — the `L` on the first operand promotes the whole left-to-right product to `long`, giving the intended 62,135,596,800,000 ms. ✔️
- **`toTimeStamp` now matches the Slice contract and both reference helpers exactly**: `(toEpochMilli() + MILLIS_BEFORE_EPOCH) * TICKS_PER_MILLISECOND` is the identical shape to JS (`js/Ice/bidir/timeUtil.ts:9`) and mathematically equivalent to C++ (`cpp/common/Time.h:44-54`, which adds the day offset before scaling, at microsecond precision). ✔️
- **`toZonedDateTime` is the exact inverse**: `timeStamp / TICKS_PER_MILLISECOND - MILLIS_BEFORE_EPOCH` mirrors C++'s `toTimePoint` (scale down, then subtract). Java-produced timestamps round-trip losslessly (whole-ms ticks); C++-produced timestamps (µs precision) truncate to the containing millisecond — correct floor behavior consistent with the helper's millisecond precision. ✔️
- **The stated verification value is right**: 62,135,596,800,000 × 10,000 = 621,355,968,000,000,000 — the well-known .NET `DateTime.UnixEpoch.Ticks`. ✔️
- **No overflow headroom concerns**: current epoch millis + offset ≈ 6.4×10¹³, ×10⁴ ≈ 6.4×10¹⁷, comfortably under `Long.MAX_VALUE`. ✔️
- **The fix is complete with this one file**: all six consumers (bidir, callback, Glacier2/callback clients and servers) route through these two functions and nothing else uses the helper; both mirror-image mistakes are fixed together, so Java↔Java pairings stay consistent while cross-language pairings become correct. Timestamps are in-flight only — nothing persisted to migrate. ✔️

### Style

Comment capitalization/punctuation tweaks match the repo's style; `719_162L` grouping is consistent with the existing `10_000`.

**Verdict: approve.** The fix is exactly right, in both directions, and closes #802 completely.

java/util/src/main/java/com/example/util/Time.java
bernardnormier: The magic number here is 719162, the days before the epoch. So the comment makes sense.

PR #824, opened 2026-08-05
Fix assorted low-severity defects in Java demos

### LINKED ISSUE #814: [Audit-Low] Java demos: assert-false interrupt handling, divergent sensor ranges, format/quote defects, misc
> **AI-generated findings** — batch of Low-severity Java items from the [2026-08-04 ice-demos correctness audit][audit] (DEMO-L03, verified) and its [verification report][verif] (NEW-L05, NEW-L09, NEW-L10, NEW-L12, NEW-L16) by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`). Code is pinned to `main` at [`99afbe3`][commit]. **Not yet human-triaged** — please verify before acting.

## 1. Interruption handling relies on disabled-by-default assertions (DEMO-L03)

Exactly 23 Java files use `assert false` in `catch (InterruptedException)` blocks (mostly shutdown-hook `mainThread.join()`; representative: [`java/Ice/greeter/client/.../Client.java#L29-L35`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Ice/greeter/client/src/main/java/com/example/ice/greeter/client/Client.java#L29-L35)); `IceStorm/weather` `Station.java:61` additionally uses it for `catch (BadQoS | AlreadySubscribed)`. Assertions are off by default, so an unexpected interruption is silently swallowed with its interrupt status cleared. Fix: `Thread.currentThread().interrupt()` + return/propagate; keep the assert only as a supplement.

## 2. Weather sensor temperature range diverges from all eight other languages (NEW-L05)

[`java/IceStorm/weather/sensor/.../Sensor.java#L63-L65`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/IceStorm/weather/sensor/src/main/java/com/example/icestorm/weather/sensor/Sensor.java#L63-L65) uses `nextInt(250, 350)` → 25.0–35.0 °C; cpp/csharp/js/python/ruby/matlab/php/swift all use 19.0–23.0 °C (humidity matches). Also in `java/Ice/optional` client1/client2. Internally consistent within Java, but mixed-language weather runs display visibly inconsistent readings.

## 3. `String.format` given .NET-style `{{` escapes (NEW-L09)

[`java/Ice/customError/client/.../Client.java#L46-L50`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Ice/customError/client/src/main/java/com/example/ice/customerror/client/Client.java#L46-L50) — `String.format` has no brace escaping, so the output literally prints `GreeterException {{ … }}` where C#/Swift print single braces.

## 4. Unbalanced quote in middleware log (NEW-L10)

[`java/Ice/middleware/server/.../AuthorizationMiddleware.java#L25`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Ice/middleware/server/src/main/java/com/example/ice/middleware/server/AuthorizationMiddleware.java#L25) — `"…invalid token " + token + "'"` is missing the opening `'` (C#/Swift print `'{token}'`).

## 5. Glacier2 session server relies on null→empty sequence marshaling (NEW-L12)

[`SharedPokeBox.java#L33-L37`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Glacier2/session/server/src/main/java/com/example/glacier2/session/server/SharedPokeBox.java#L33-L37) returns `retrieveCollection(...)`, which returns `null` for a missing collection; the same class null-checks in `caught` but not here. Not a crash — `slice2java`-generated marshaling writes null sequences as size 0 (verified) — but it contradicts the Slice `@return` contract and the explicit `?? []` coalescing in the C#/Swift counterparts.

## 6. Misc (NEW-L16 d–g)

- `java/IceDiscovery/replication` — the three files under `src/main/java/com/example/icediscovery/greeter/...` declare `package com.example.icediscovery.replication...`; package and directory disagree (Gradle tolerates it, IDEs and sourcepath tooling don't).
- [`java/IceDiscovery/replication/server/.../Server.java#L45`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/IceDiscovery/replication/server/src/main/java/com/example/icediscovery/greeter/server/Server.java#L45) takes the *first* 4 UUID chars; C#/Swift take the *last* 4.
- [`java/Ice/config/server/.../Server.java#L35`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Ice/config/server/src/main/java/com/example/ice/config/server/Server.java#L35) hardcodes "Listening on port 4061" although the endpoint comes from `server.conf` (C#/Swift deliberately print nothing and defer to `Ice.Trace.Network`).
- `java/Glacier2/session/server/.../Server.java:47` is the repo's only empty `catch (InterruptedException ignored) {}` (all other servers use the commented `assert false` pattern — see item 1).

bernardnormier:
Fixes #814 (items 2-6; item 1 is deliberately not addressed, see below).

- **Sensor temperature range** (item 2): `IceStorm/weather` sensor and both `Ice/optional` clients now generate 19.0-22.9 °C like the other eight language ports (humidity was already aligned).
- **customError client** (item 3): `String.format` has no brace escaping, so the .NET-style `{{ }}` printed literal double braces; now prints `{ }` like C#/Swift.
- **middleware server** (item 4): added the missing opening quote in the "Rejecting request with invalid token" log.
- **Glacier2/session SharedPokeBox** (item 5): `getInventory` now returns an empty list instead of null for a missing collection, honoring the Slice contract like the C#/Swift `?? []`.
- **IceDiscovery/replication** (item 6): source directories renamed `greeter` → `replication` to match the declared packages (pure renames), and the greeter name now uses the last 4 UUID chars like C# (`uuid[^4..]`) and Swift.
- **Ice/config server** (item 6): the hardcoded "Listening on port 4061..." message (the endpoint actually comes from `server.conf`) is replaced with the same `Ice.Trace.Network` comment the C# port uses.
- **Glacier2/session server** (item 6): the repo's only empty `catch (InterruptedException) {}` now uses the standard commented `assert false` pattern.

**Item 1 (assert-false interrupt handling, 23 files) is intentionally not addressed:** the `assert false` in these catch blocks is documentation of an impossible path, not a runtime check — replacing it with interrupt-restore boilerplate across 23 files would add robustness plumbing that teaches nothing about Ice, contrary to the demos' keep-it-simple goal.

All seven touched demos build cleanly.


pepone (approved):
> **AI-generated review** — by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), cross-checking each change against the sibling language ports and the repo at the PR base.

### Correctness — all changes verified against the sibling ports

- **Temperature ranges** (`Sensor.java`, `optional/client1+2`): `nextInt(190, 230)` / `nextInt(230 - 190) + 190` produce 19.0–22.9 °C, which matches C# exactly (`RandomNumberGenerator.GetInt32(190, 230)` is also exclusive-bound → 19.0–22.9) and Ruby; PHP/MATLAB/Python are inclusive to 23.0 and C++ is continuous, so exact upper-bound behavior already varies by port. The PR description's "19.0-22.9" is precise. Humidity correctly untouched. ✔️
- **`customError` braces**: `{{ }}` → `{ }` is right — `String.format` has no brace escaping. ✔️
- **`AuthorizationMiddleware` quote**: now `'{token}'`, matching C#/Swift. ✔️
- **`SharedPokeBox.getInventory`**: coalescing to `Collections.emptyList()` is safe — `Collections` is already imported (used by `releaseAll`), and the returned immutable list marshals fine. Honors the Slice `@return` contract like C#'s `?? []`. Wire-compatible: the generated marshaling already wrote null sequences as size 0, so this makes the source honor the contract explicitly with no behavior change. ✔️
- **`IceDiscovery/replication` renames**: all three files move under `com/example/icediscovery/replication/`, packages were already declared `…replication…`, both `build.gradle.kts` `mainClass` entries already point at the `replication` classes, and no stale references to the old directories remain (grep clean). Pure renames (100% similarity), so `gradle run` is unaffected. ✔️
- **UUID suffix**: `substring(uuid.length() - 4)` matches C# `uuid[^4..]` and Swift `.suffix(4)`. ✔️
- **Config server**: the replacement comment is byte-identical to the C++/C#/Swift ports' comment, `server.conf` already sets `Ice.Trace.Network=1` so listening info remains visible, and no README references the removed "Listening on port 4061..." string (grep clean). ✔️
- **Empty catch → `assert false` pattern**: matches the repo-standard block (same comment wording, `catch (InterruptedException e)` as in every sibling server) and is consistent with the PR's own position on item 1. ✔️

### Suggestions (minor, optional)

- **Adjacent inconsistency in `customError/Client.java`**: the `DispatchException` branch two lines above the changed one prints `"Failed to create a greeting for '%s': '%s' '%s'"` — no exception-type framing or field names — while C# prints `DispatchException { Message = '…', replyStatus = … }` (Python likewise). Since this file is already being touched, aligning that branch would finish the file in one pass. Not part of #814, so fine to defer.
- **Issue-closing semantics**: "Fixes #814" will auto-close the whole issue including item 1. That's acceptable since the PR body documents the wontfix rationale — just confirm that's the intent rather than leaving item 1 tracked.

### On the item-1 rationale

Declining the 23-file interrupt-restore sweep is defensible: the catches sit in shutdown-hook `join()` paths where interruption genuinely cannot occur today, the comments say so, and the demos optimize for teaching Ice rather than defensive-threading boilerplate. Converting the one *empty* catch to the commented `assert false` form (rather than the reverse) makes the codebase internally consistent with that stance.

### Risk / tests

Low risk: all changes are output-formatting or range tweaks except `getInventory`, which is wire-compatible as noted above. The repo has no automated demo tests; the author reports all seven touched demos build cleanly, which is the available bar.

**Verdict: approve.** Everything in the diff is correct and convention-consistent; the two notes above are optional polish.

bernardnormier:
Thanks for the review. Both suggestions addressed:

- The `DispatchException` branch in `customError/Client.java` is now aligned with the C# port's framing (f6454d48f): `DispatchException { message = '…', replyStatus = … }`, with lowercase field names matching the file's `GreeterException` branch.
- Issue-closing semantics: closing all of #814 is intentional — item 1 is declined as documented in the PR description, so nothing remains to track.

Also renamed the unused exception variable to `ignored` per Copilot's comment (cc8198ff5).

PR #831, opened 2026-08-05
Fix Python Glacier2 session client to rethrow unexpected dispatch errors

bernardnormier:
Fixes #807.

The client's final assertion caught every `Ice.DispatchException` but printed only when `replyStatus == Unauthorized` — any other status was silently swallowed and the process exited 0, making a failed run look successful. Python was the only language port without the rethrow; this adds `else: raise`, mirroring the C++ shape.

Verified end-to-end against the nightly (glacier2router + Python server + client): the full demo flow runs through the fixed catch and prints "The PokeBox proxy remains unusable, as expected." with exit 0. `ruff check` and `ruff format --check` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #833, opened 2026-08-05
Fix cross-language divergences in the Glacier2 session demo

### LINKED ISSUE #819: [Audit-Low] Glacier2 session demo cross-language divergences (C# catch-count off-by-one, dead assignments, output polish)
> **AI-generated findings** — batch of Low-severity cross-language divergences in the Glacier2 session demo, from the [verification report][verif] (NEW-L03, NEW-L14) accompanying the [2026-08-04 ice-demos correctness audit][audit], by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`). Code is pinned to `main` at [`99afbe3`][commit]. **Not yet human-triaged** — please verify before acting.

## 1. C# catch count is 1–5 where every other language is 1–6 (NEW-L03)

https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/csharp/Glacier2/Session/Client/Program.cs#L48-L50

`RandomNumberGenerator.GetInt32(1, 6)` has an exclusive upper bound. C++ (`uniform_int_distribution{1, 6}`), Java (`nextInt(6) + 1`), Python (`randint(1, 6)`), and Swift (`Int.random(in: 1...6)`) all yield 1–6, so the C# run can never catch 6 and reaches the "all the Pokémon escaped" branch more slowly.

## 2. Cross-language polish (NEW-L14)

- **Dead assignment in four languages:** the re-created session's return value is assigned and never used in C++ (`Client.cpp:115`), C# (`Program.cs:92`), Python (`main.py:137`), and Java (`Client.java:106`); only Swift validates it.
- **Java exits 0 on session-creation failure:** [`Client.java#L49-L54`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/java/Glacier2/session/client/src/main/java/com/example/glacier2/session/client/Client.java#L49-L54) (and L105-L110) prints and returns, where every other language fails loudly — combined with the checked-exception path, Java is the only language with a "success" exit that never reaches the final assertion.
- **Java prints "Pokemon"** in user-visible output while its own comment (line 67) and all sibling ports use "Pokémon".
- **C# doesn't terminate the "Catching…" line:** `Program.cs:50` uses `Console.Write`, so the next output is appended to the same line (all siblings print a newline).
- `python/Glacier2/session/client/main.py:121` says "Glacier router" (siblings: "Glacier2 router").

bernardnormier:
Fixes #819.

Swift was already correct on every point; the other four ports converge on it:

- **C# catch count**: `RandomNumberGenerator.GetInt32(1, 6)` has an exclusive upper bound, so C# could never catch 6 Pokémon; now `GetInt32(1, 7)`, matching the 1–6 of C++/Java/Python/Swift.
- **Recreated session validated** in C++ (`assert(session)`), C# (`Debug.Assert`), Java (`assert`), and Python (`assert`) instead of a dead assignment, mirroring Swift's guard and each port's own first-session validation style.
- **Java fails loudly on session-creation failure**: both print-and-return catch blocks removed; `main` declares the checked Glacier2 exceptions, so a failure exits non-zero like every other port (net −10 lines).
- **Java prints "Pokémon"** in its four output lines, consistent with its own comments and the siblings. (The inventory listing already prints `Nidoran♀`/`Nidoran♂` in all ports, so this adds no new non-ASCII exposure.)
- **C#** terminates the "Catching…" line with `WriteLine`.
- **Python** comment typo: "Glacier router" → "Glacier2 router".

Verified: Python `py_compile` + ruff clean, C++ demo builds against the nightly, C# `dotnet build` 0 errors, Java `:client:build` green.

Note: touches `python/Glacier2/session/client/main.py` in a different hunk than #831 — no conflict.

### DIFF
--- a/java/Glacier2/session/client/src/main/java/com/example/glacier2/session/client/Client.java
+++ b/java/Glacier2/session/client/src/main/java/com/example/glacier2/session/client/Client.java
@@ -27,7 +27,7 @@ class Client {
         "Primeape"
     };
 
-    public static void main(String[] args) {
+    public static void main(String[] args) throws PermissionDeniedException, CannotCreateSessionException {
 
         // Retrieve the user ID for this run.
         final String userId = args.length > 0 ? args[0] : System.getProperty("user.name");
@@ -45,13 +45,7 @@ class Client {
             // Create a session with the Glacier2 router. In this demo, the Glacier2 router is configured to accept any
             // username/password combination. This call establishes a network connection to the Glacier2 router; the
             // lifetime of the session is the same as the lifetime of the connection.
-            SessionPrx session;
-            try {
-                session = router.createSession(userId, "password");
-            } catch (PermissionDeniedException | CannotCreateSessionException e) {
-                System.out.println("Could not create session: " + e.getMessage());
-                return;
-            }
+            SessionPrx session = router.createSession(userId, "password");
 
             // We configured a SessionManager on the Glacier2 router, so 'session' is a non-null 'PokeSession'.
             PokeSessionPrx pokeSession = PokeSessionPrx.uncheckedCast(session);
@@ -62,12 +56,12 @@ class Client {
             assert pokeBox != null;
 
             int currentCount = pokeBox.getInventory().size();
-            System.out.println(userId + "'s PokeBox contains " + currentCount + " Pokemon.");
+            System.out.println(userId + "'s PokeBox contains " + currentCount + " Pokémon.");
 
             // Catch a few Pokémon.
             var randomSource = new java.util.Random();
             int addCount = randomSource.nextInt(6) + 1;
-            System.out.println("Catching " + addCount + " Pokemon... ");
+            System.out.println("Catching " + addCount + " Pokémon... ");
             List<String> newPokemon = randomSource.ints(addCount, 0, ALL_POKEMON.length)
                 .mapToObj(i -> ALL_POKEMON[i])
                 .toList();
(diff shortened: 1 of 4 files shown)


java/Glacier2/session/client/src/main/java/com/example/glacier2/session/client/Client.java
```diff
@@ -62,26 +56,26 @@ public static void main(String[] args) {
             assert pokeBox != null;
 
             int currentCount = pokeBox.getInventory().size();
-            System.out.println(userId + "'s PokeBox contains " + currentCount + " Pokemon.");
+            System.out.println(userId + "'s PokeBox contains " + currentCount + " Pokémon.");
```
InsertCreativityHere, line 59: For the record, on Windows 11 25H2, with Java 21 LTS, this still doesn't display correctly.
In stock Powershell:
Image
and in the older Command Prompt:
Image

InsertCreativityHere: To be safe, I installed the 25 LTS, with the latest Powershell as well, interestingly, this made the text even more broken!
Image
Also spat out some unrelated warnings... We might need to update our demos to work better with JDK25+

InsertCreativityHere (approved):
Looks good to me!

But we should keep the accented 'e' character out of the Java version of the demo.
I checked to see if there was a simple flag or setting we could tweak to make this work on Windows, and there is not.

java/Glacier2/session/client/src/main/java/com/example/glacier2/session/client/Client.java
pepone: Where are the warnings coming from? seems could be related to the gradle version we use.

InsertCreativityHere: The warnings are coming from gradle. Running the demo directly with Java, no warnings.
Not clear if it's just we need to update gradle, or if there's some setting or something we need to tweak.
Either way, I opened a separate issue for this. Not related to this PR.

bernardnormier: For the é display, see #868.

pepone (approved):

PR #838, opened 2026-08-06
Use catch-where clauses in Swift Middleware and Glacier2 Session clients

bernardnormier:
Simplifies the expected-error handling in two Swift demo clients by replacing catch blocks that test the reply status and rethrow with `catch ... where` clauses:

- **Ice/Middleware client**: the catch now matches only the expected `unauthorized` reply status; any other error propagates naturally instead of being rethrown explicitly.
- **Glacier2/Session client**: same pattern for the expired-session check on `getInventory`.

Behavior is unchanged. Both demos build cleanly and pass `swift format lint`.

### DIFF
--- a/swift/Glacier2/Session/Sources/Client/main.swift
+++ b/swift/Glacier2/Session/Sources/Client/main.swift
@@ -96,11 +96,9 @@ do {
     // The pokeBox proxy no longer works as it was created with the token of an old session.
     _ = try await pokeBox.getInventory()
     print("Error: the PokeBox proxy should not work with a new session!")
-} catch let error as Ice.DispatchException {
-    if error.replyStatus == ReplyStatus.unauthorized.rawValue {
-        // See code in SharedPokeBox.getUserId.
-        print("The PokeBox proxy remains unusable, as expected.")
-    } else {
-        throw error
-    }
+} catch let error as Ice.DispatchException
+    where error.replyStatus == ReplyStatus.unauthorized.rawValue
+{
+    // See code in SharedPokeBox.getUserId.
+    print("The PokeBox proxy remains unusable, as expected.")
 }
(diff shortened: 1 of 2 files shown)


pepone (approved):
> **AI-generated review** — by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), verifying semantics and coverage against current `main` and the sibling language ports.

### Correctness — verified

- **Semantics are identical.** A `catch let error as DispatchException where error.replyStatus == ReplyStatus.unauthorized.rawValue` clause matches only when both the cast and the condition hold; any other error is simply never caught, and in top-level `main.swift` code that terminates the client loudly — exactly where the previous explicit `throw error` also ended up. Expected status → print; anything else → loud failure. Unchanged in both files. ✔️
- **Coverage is complete**: these two are the only catch-then-test-and-rethrow sites in `swift/` — nothing with the old pattern is left behind. ✔️
- **Cross-language convention improves**: this is the Swift analog of the C# exception filters the sibling clients already use (`catch … when (… == Ice.ReplyStatus.Unauthorized)` in `csharp/Glacier2/Session/Client/Program.cs:100`), so the two mappings now express the expected-error idiom the same way, while C++/Java/Python keep their explicit-rethrow shape. ✔️
- The teaching point is preserved: unexpected reply statuses still fail loudly rather than being swallowed. ✔️

### Minor (optional)

- The new Middleware comment says "any other error propagates **to the caller**" — the exception has already crossed the wire by this point, and top-level code has no caller; an unmatched error just ends the client process. Suggested wording:

  ```swift
  // An Unauthorized error is expected with an invalid (or missing) token; any other error is not caught here
  // and terminates the client program. See AuthorizationMiddleware.
  ```

**Verdict: approve.** Correct, complete, and a genuine readability improvement that also aligns Swift with the C# filter idiom.

bernardnormier:
> ### Minor (optional)
> * The new Middleware comment says "any other error propagates **to the caller**" — the exception has already crossed the wire by this point, and top-level code has no caller; an unmatched error just ends the client process. Suggested wording:
>   ```swift
>   // An Unauthorized error is expected with an invalid (or missing) token; any other error is not caught here
>   // and terminates the client program. See AuthorizationMiddleware.
>   ```

I prefer the existing wording with "caller" - meaning the client code.

PR #858, opened 2026-08-06
Explain the hosted hello.zeroc.com default in client-only Greeter READMEs

### LINKED ISSUE #847: Explain the hosted hello.zeroc.com default in client-only Greeter READMEs
Split out from #791 (theme A5) — editorial follow-ups from the internal demo-review report.

For the client-only languages (`js`, `matlab`, `php`, `ruby`), the Greeter README should explain that the client defaults to the hosted `hello.zeroc.com` Greeter server, and how to switch the proxy string to `localhost` for a local server.

Note: the second part of this theme (point other client-only demos at a concrete matching server demo path rather than just naming candidate languages) was already done in #834.

bernardnormier:
The js, matlab, php, and ruby Greeter clients connect by default to the Greeter server hosted by ZeroC at `hello.zeroc.com`, but their READMEs didn't say so — a reader could think they must first start a server, when the demo works out of the box.

This PR extends the existing `> [!NOTE]` block in each README:

> Ice for PHP supports only client-side applications. By default, the client connects to a Greeter server hosted by ZeroC at `hello.zeroc.com`; as a result, you don't need to start a Greeter server to run this demo. If you want to use your own Greeter server (implemented in a language with server-side support, such as C++, C#, Java, Python, or Swift), replace `hello.zeroc.com` with `localhost` in the proxy string in `Client.php`.

The four files get the same wording, varying only the language name and the client file name (`client.ts`, `client.m`, `Client.php`, `client.rb`). The "replace hello.zeroc.com" instruction matches the comment already present in each client's source.

Verified against the hosted server: the js client, built and run with no local server, prints the greeting.

Fixes #847.

### DIFF
--- a/js/Ice/greeter/README.md
+++ b/js/Ice/greeter/README.md
@@ -26,5 +26,7 @@ node client.js
 ```
 
 > [!NOTE]
-> Ice for JavaScript has limited server-side support. In this demo, we connect to a Greeter server implemented in a
-> language with full server-side support, such as C++, C#, Java, Python, or Swift.
+> Ice for JavaScript has limited server-side support. By default, the client connects to a Greeter server hosted by
+> ZeroC at `hello.zeroc.com`; as a result, you don't need to start a Greeter server to run this demo. If you want to
+> use your own Greeter server (implemented in a language with full server-side support, such as C++, C#, Java, Python,
+> or Swift), replace `hello.zeroc.com` with `localhost` in the proxy string in `client.ts`.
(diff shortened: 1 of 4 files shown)


pepone (approved):
Looks good overall. I verified that the JavaScript client works with the hosted server and returns the expected greeting. One minor wording issue applies to all four READMEs: 'use your own Greeter server' is broader than the instruction to replace hello.zeroc.com with localhost, which only works when the server runs on the same computer. Could we say: 'For a local Greeter server, replace hello.zeroc.com with localhost; for a server running on another computer, use its hostname or IP address.' This would also match the existing comments in the client source files.

bernardnormier:
Reworded in f691f3d (branch also rebased on current main). All four NOTEs now end with:

> edit the proxy string in `Client.php`: replace `hello.zeroc.com` with `localhost` for a server running on the same computer, or with the server's hostname or IP address for a server running on another computer.

(with the per-language client file name), which also matches the existing comments in the client source files.

Issue #864, opened 2026-08-06
Switch the Java Config demo from JKS to PKCS12 keystores


bernardnormier:
The `java/Ice/config` demo is the last user of the legacy JKS keystore format: both conf files point at the `.jks` files in the shared `certs/` directory.

- `java/Ice/config/client/client.conf` — `IceSSL.Keystore=client.jks`, `IceSSL.Truststore=ca.jks`
- `java/Ice/config/server/server.conf` — `IceSSL.Keystore=server.jks`, `IceSSL.Truststore=ca.jks`

JKS is a proprietary format that `keytool` itself recommends migrating away from ("The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12"); PKCS12 has been the JVM's default keystore type since Java 9.

`certs/` already contains the PKCS12 equivalents (`client.p12`, `server.p12`, `ca.p12`) used by the other languages and by `java/Ice/secure`, so the fix is to point the two conf files at the `.p12` files instead (with `IceSSL.KeystoreType=PKCS12` if needed). Once done, `ca.jks`, `client.jks`, and `server.jks` can be removed from `certs/`.

This is also a step toward #862: going password-less in Java requires leaving JKS behind anyway, since `keytool` refuses to create a JKS keystore with a password shorter than 6 characters.

PR #865, opened 2026-08-06
Refresh expired SSL test certificates

bernardnormier:
The SSL test certificates in `certs/` **expired on June 25, 2026**, so every demo that runs with `Ice.Default.Protocol=ssl` (the cpp/csharp/java/python/swift Config demos) has been failing certificate validation since then.

`certs/` is a copy of `ice/certs/common/ca`. The ice repository regenerated its certificates the day they expired (zeroc-ice/ice#5705, regenerated again by zeroc-ice/ice#6024 to strip the PKCS12 friendlyName), but the copy here was never synced. This PR syncs all 13 files (`.pem`, `.der`, `.p12`, `.jks`) from current ice main; the new certificates are valid until **July 28, 2027**.

Verified: the python Config demo completes a greet request over `transport = ssl` with the refreshed certificates (it fails peer verification with the old ones).

Note for maintainers: these certificates are generated with a 398-day lifetime (macOS requirement), so this same breakage will recur around July 2027 unless the sync from ice happens with each regeneration.

Found while working on #864, which will follow as a separate PR on top of this one.


pepone (approved):
Approving. I verified this mechanically rather than by eye — it is a faithful upstream sync and it unblocks every
SSL demo. Nothing blocking.

One caveat worth knowing about before this lands: on macOS the refreshed certificates do not displace the old ones
in an existing keychain, so returning developers will need a one-line cleanup. That is a pre-existing property of
how the certificates are generated, not something this PR introduced, and the durable fix is tracked separately in
#869 — details in §1 below. Everything after that is small suggestions, take them or leave them.

## What I checked

Binary blobs are unreviewable by eye, so I checked them mechanically:

- **Provenance.** All 13 files are byte-identical to `zeroc-ice/ice` `certs/common/ca` — on ice `main` *and* on ice
  `3.8` (compared git blob SHAs, not just sizes). This is a faithful sync, not a local regeneration.
- **Internal consistency.** For ca/client/server the certificate public key matches the corresponding `*_key.pem`;
  `client_cert.pem` and `server_cert.pem` verify against `ca_cert.pem`; `ca_cert.der` is the DER form of
  `ca_cert.pem`. Subjects, SANs (`DNS:localhost`, `IP:127.0.0.1`), key usages and EKUs are unchanged.
- **Passwords and aliases.** `password` still opens every `.p12` and `.jks`, and still unlocks the keys (checked via
  `KeyManagerFactory.init`, which validates the key password — `keytool -list` does not). JKS aliases `ca` / `client`
  / `server` are preserved.
- **Red → green.** A JSSE mutual-TLS handshake with hostname verification against `localhost` succeeds with the new
  `.jks` and `.p12`, and fails with the old ones (`CertificateExpiredException: NotAfter: Thu Jun 25 ... 2026`).
- **Backport.** `certs/` on `origin/3.8` is byte-identical to `certs/` on `main` before this PR, so 3.8 is equally
  broken and needs this. I cherry-picked `b3956f7` onto `origin/3.8` in a throwaway worktree: applies cleanly, no
  conflicts.

CI being green here is not evidence of anything: the workflows only build and lint, and never run a demo, so `main`
stayed green for the whole ~6 weeks the certificates were expired.

## 1. macOS: the new certificates reuse the old serial numbers, so a stale keychain wins (non-blocking)

The regenerated leaves keep the serial numbers of the ones they replace — client `0x1000`, server `0x1001` — under
an unchanged issuer DN. The macOS keychain keys certificate records on (type, issuer, serial), so in any keychain
that still holds the old certificate the new one is a *duplicate* and is silently dropped. Different Subject Key
Identifiers do not help; that is not part of the keychain's uniqueness rule.

Verified with a throwaway keychain (the login keychain was not touched):

```shell
# a keychain that already holds the old server certificate — i.e. any developer who ran the demo before June 25
security import old/server.p12 -k demo.keychain -P password   # "1 identity imported. 1 certificate imported."
security import new/server.p12 -k demo.keychain -P password   # "2 certificates imported."  <- no identity
security find-identity demo.keychain
#  1) A822CF4C... "ca.server" (CSSMERR_TP_CERT_EXPIRED)   <- still the expired one, the new leaf never landed
```

A direct `SecItemAdd` of the new server certificate into that keychain returns `errSecDuplicateItem` (-25299), while
the new *CA* certificate — whose serial did change — adds fine.

Who this bites, concretely:

- **`cpp/Ice/secure` on macOS.** The README tells the user to `security import ../../../certs/server.p12`, and the
  server then picks its identity with `kSecMatchSubjectContains "ca.server"` + `kSecMatchLimitOne`
  (`cpp/Ice/secure/UtilSecureTransport.cpp`). Anyone who ran this demo before the expiry re-runs the documented
  import, gets no new identity, and the server keeps serving the expired certificate — so the demo still fails and
  it looks like the fix did not work.
- **The Config demos that set `IceSSL.Keychain`** (cpp, swift, python, php, ruby) write `client.keychain` /
  `server.keychain` next to the demo, and those files are gitignored, so they survive a `git pull` unnoticed. With
  `IceSSL.CertFile=<x>.p12` and no `IceSSL.KeyFile`, Ice imports through `SecItemImport`
  (`SecureTransportUtil.cpp`, the `keyFile.empty()` branch), which is exactly what `security import` does above — so
  the stale keychain keeps the old certificate. I did not run these demos, so I have not seen the resulting Ice-side
  error text; the keychain behaviour itself is verified.

Windows and Linux are unaffected: they load the certificate from the file on every run rather than from a store that
persists it across runs.

Immediate workaround for anyone hitting this:

```shell
security delete-identity -c ca.server            # cpp/Ice/secure, then re-run the documented import
rm -f *.keychain                                 # in the Config demo directory
```

Filed as #869 for the Config demos: since zeroc-ice/ice#5343, leaving `IceSSL.Keychain` unset makes the macOS
transport import into a temporary keychain that is deleted with the communicator, so dropping the property removes
this failure mode for good. Its cherry-pick to the ice-demos `3.8` branch has to wait until zeroc-ice/ice#5343 is on
the Ice `3.8` branch, after which the macOS Config demos there require Ice >= 3.8.3.

`cpp/Ice/secure` manages its own keychain, so #869 does not cover it — filed as #870. That one turns out to matter
beyond this PR: the demo imports into the login keychain, which per zeroc-ice/ice#5343 cannot complete an ECDHE
handshake on the server side, so it should be showing a dedicated keychain regardless of the certificate refresh.

The root cause of the collision is upstream: `certs/makecerts.sh:38` resets the serial file to `1000` on every
regeneration, so every future refresh collides the same way. Filed as zeroc-ice/ice#6488, together with a request
that the script print a reminder to sync the copy in this repository — which is the step that was missed after
zeroc-ice/ice#5705 and zeroc-ice/ice#6024 and left the demos broken for six weeks.

## 2. `cpp/Ice/secure/keychain-access.png` is now stale (non-blocking)

The screenshot reads *server wants to sign using key "server" in your keychain*. That name came from the PKCS#12
`friendlyName`, which zeroc-ice/ice#6024 stripped. Importing the new `server.p12` now produces a generated hex name
instead — and it is not even stable across imports (I got `A08297C5`, `A9FEE10E`, `18BACF39` on three runs). A user
following the README will see a dialog naming an unrecognizable key and reasonably wonder whether it is the right
one. Worth a caption tweak, or dropping the key name from the surrounding prose.

Everything else about the friendlyName strip is harmless here: nothing in this repository looks a certificate up by
alias or friendly name. (For the record, Java now derives alias `1` for the `.p12` files instead of `ca` / `client` /
`server`; the `.jks` aliases are unchanged.)

## 3. `certs/README.md` should carry the expiry date and the re-sync step (suggestion)

The recurrence warning — 398-day lifetime, so this breaks again around July 2027 — is in the PR description, which
nobody will be reading in 2027. `certs/README.md` currently says only that the certificates are "copied from the main
Ice repository". One line recording the current expiry (2027-07-28) and "re-copy all 13 files from `ice/certs/common/ca`
whenever ice regenerates them" would put it where the next person looks.

## 4. A CI guard (author's call, separate PR)

Nothing in CI touches these files, which is why this went unnoticed. A ~5-line job looping
`openssl x509 -checkend $((30*86400)) -noout -in <cert>` over `certs/*_cert.pem` would turn the next occurrence into
a CI failure 30 days ahead of the breakage instead of a silently broken demo. Not this PR's job, but cheap.

## 5. Description nit

"the cpp/csharp/java/python/swift Config demos" undercounts it — the matlab, php and ruby Config clients also set
`Ice.Default.Protocol=ssl`, and the cpp, csharp and java Secure demos consume these files too. Worth correcting in the
commit message since it is the record of what was broken.

Issue #862, opened 2026-08-06
Switch the Config demos to password-less certificates


bernardnormier:
The Config demos use password-protected certificates: the `.conf` files set `IceSSL.Password=password` (and `IceSSL.KeychainPassword=password` where SecureTransport applies) in:

- `cpp/Ice/config/client.conf` and `server.conf`
- `csharp/Ice/Config/Client/client.conf` and `Server/server.conf`
- `java/Ice/config/client/client.conf` and `server/server.conf`
- `matlab/Ice/config/client.conf`
- `php/Ice/config/client.conf`
- `python/Ice/config/client/client.conf` and `server/server.conf`
- `ruby/Ice/config/client.conf`
- `swift/Ice/Config/client.conf` and `server.conf`

Switching to password-less certificates would drop these properties and simplify the demo configuration. The certificates under `certs/` are copied from the main Ice repository, so they would need to be regenerated there (or new password-less test certificates copied in) first.

Note for Java: the demo currently uses `.jks` keystores, and `keytool` refuses to create a keystore with a password shorter than 6 characters. Going password-less there means switching the demo to PKCS12 keystores (the JVM default keystore type since Java 9).

This replaces the `# TODO: switch to password-less certificates` comments that used to sit in the C++, C#, Python, and Swift `.conf` files (removed as part of #855).

externl:
IIRC there's an issue doing this in C++ in the macOS APIs that use our generated pkcs12 require a password.

PR #866, opened 2026-08-06
Switch the Java Config demo from JKS to PKCS12 keystores

### LINKED ISSUE #864: Switch the Java Config demo from JKS to PKCS12 keystores

bernardnormier:
Switches the `java/Ice/config` demo — the last JKS user in the repo — to the PKCS12 keystores already shipped in `certs/`, and deletes `ca.jks`, `client.jks`, and `server.jks` since nothing references them anymore. JKS is a legacy proprietary format; `keytool` itself recommends migrating to PKCS12, the JVM default keystore type since Java 9.

Two things worth calling out in the conf diff:

- **`IceSSL.KeystorePassword` / `IceSSL.TruststorePassword` are new — and required.** Unlike JKS (which only uses the store password for an integrity check), our PKCS12 files encrypt the certificate bags with the store password (`makecerts.sh` exports with `-certpbe PBE-SHA1-3DES` for macOS compatibility). Without these properties the keystore silently loads with a private key but **no certificate chain**, and every handshake fails with `SSL_NULL_WITH_NULL_NULL` / "error while encoding message". Filed zeroc-ice/ice#6485 for the stale IceSSL doc on this.
- The client.conf truststore comment said "the keystore the **server** uses" — fixed to "client".

Verified end-to-end on macOS/JDK 17: server and client greet over `ssl` with a TLS cipher in the security trace. Note the verification requires the refreshed certificates from #865 — the expired ones in main fail validation regardless of keystore format (CI runs no demos, so CI is green either way).

If #862 (password-less certificates) happens, all three password properties disappear from these confs.

Fixes #864.


pepone (approved):
Reviewed and tested end to end together with #865. I cherry-picked #865 followed by this PR, accepted this PR's deletion of the three refreshed JKS files, and built the Java Config demo successfully. The client completed the greet request, and the server trace confirmed TLS 1.3 using TLS_AES_256_GCM_SHA384. I found no issues with the implementation. #865 should merge first, after which this branch will need updating to resolve the expected JKS modify/delete conflicts. With that sequencing handled, LGTM.

Issue #868, opened 2026-08-06
Proposal: run all Java demos via installDist launchers instead of `gradlew run`


bernardnormier:
### Proposal

Standardize every Java demo README on the application plugin's generated launchers — `gradlew installDist`, then `bin/<app>` (`bin\<app>.bat` on Windows) — for **both servers and clients**, and drop `gradlew :server:run` / `gradlew :client:run` from the instructions.

### Why

`gradlew run` launches the application JVM as a child of the Gradle **daemon**, detached from the terminal (separate process group, stdout piped through the daemon). That single fact causes three separate failures, all at the Windows/Gradle seam, none of them in the Ice code being demonstrated:

1. **Ctrl+C** doesn't reach the app JVM — shutdown hooks don't display and the process lingers holding its port. (#413)
2. **Non-ASCII console output** mangles — the piped stdout forces `stdout.encoding` to the ANSI code page regardless of `chcp`, so e.g. `Pokémon`/`Nidoran♀` render as `Pok?mon`/mojibake. This affects **clients**, not just servers.
3. **`./gradlew`** in the current instructions doesn't run on Windows **cmd.exe** at all (`'.' is not recognized`), so the instructions were never actually cross-platform-uniform to begin with.

Running the installed launcher attaches the app JVM to the real terminal, fixing 1 and 2; and since we're documenting per-platform commands anyway (3), the launcher's OS-specific path is no longer a regression in uniformity — the uniformity was illusory.

### Why everywhere (not servers-only)

A servers-only fix (installDist for servers, keep `gradlew run` for clients) assumes the relevant axis is the app's *role*. It isn't. The axis is **"does this program need a real terminal"** — for signal delivery (Ctrl+C), correct console I/O encoding, or interactive stdin. Two demos make this concrete:

- **Glacier2 session** — the *client* prints `Pokémon`/`Nidoran♀`, which mangles under `gradlew :client:run` (symptom 2). So clients are not universally safe on the wrapper.

- **IceStorm weather** — this demo has **no plain server and no run-and-exit client**. The `sensor` is a publisher you Ctrl+C; the `station` is a subscriber you Ctrl+C (`communicator.waitForShutdown()`). Both print `"Press Ctrl+C to stop..."` / `"Caught Ctrl+C, ..."` and implement graceful-shutdown hooks — the whole point of the demo. Under `gradlew :sensor:run` / `gradlew :station:run`:
[...]

PR #872, opened 2026-08-09
Run the Java Greeter demos via installDist launchers

bernardnormier:
Implements the proposal in #868 for the Java `Ice/greeter` and `IceGrid/greeter` demos.

- `Ice/greeter`: build with `./gradlew installDist` and run the applications via the generated launcher scripts instead of `gradlew run`.
- `IceGrid/greeter`: run the client via its launcher script (the build already used `installDist`).
- Instead of separate Linux/macOS and Windows command flavors, each README keeps a single set of commands with a note telling Windows users to run them in Git Bash or PowerShell.

Both demos were tested on Windows (PowerShell and Git Bash) and the commands work as documented, including clean Ctrl+C shutdown of the servers.

### DIFF
--- a/java/Ice/greeter/README.md
+++ b/java/Ice/greeter/README.md
@@ -9,6 +9,9 @@ This demo provides two implementations for the server:
 
 The same client works with both.
 
+> [!NOTE]
+> On Windows, run all the commands below in Git Bash or PowerShell; they don't work in the cmd.exe Command Prompt.
+
 ## Building the demo
 
 The demo has three Gradle projects, **client**, **server**, and **serveramd**, all using the [application plugin].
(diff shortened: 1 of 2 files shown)


java/Ice/greeter/README.md
```diff
@@ -9,34 +9,40 @@ This demo provides two implementations for the server:
...
 The demo has three Gradle projects, **client**, **server**, and **serveramd**, all using the [application plugin].
 
 To build the demo, run:
 
 ```shell
-./gradlew build
+./gradlew installDist
 ```
 
+This creates a self-contained distribution under build/install/ for each application, with launcher scripts in its
```
InsertCreativityHere, line 25: Fine to keep this, but also fine to remove it. It’s just explaining how the ‘installDist’ command works, which is helpful, but arguably outside the scope of demo’a README.

InsertCreativityHere (approved):
Looks good to me!

Can confirm, it fixes the CTRL-C issue completely.
We should do the same for all the demos after this PR.

pepone (approved):

externl (approved):

PR #873, opened 2026-08-10
Run all Java demos via installDist launchers

### LINKED ISSUE #413: Ctrl+C with Java servers
When you follow the instructions in the README and Ctrl+C a server, you should see the shutdown message from the shutdown hook. Currently, nothing is displayed.

### LINKED ISSUE #868: Proposal: run all Java demos via installDist launchers instead of `gradlew run`

bernardnormier:
Implements the proposal in #868 for all remaining Java demos, following the model established in #872 for the Greeter demos:

- Build with `./gradlew installDist` and run the applications via the generated launcher scripts instead of `gradlew run`, unwrapping `--args="…"` into plain arguments.
- Each README keeps a single set of commands with a note telling Windows users to run them in Git Bash or PowerShell — the "not all of them work in cmd" variant for demos that also run Ice services (`glacier2router`, `icegridregistry`, `icebox`), which do work in cmd.
- In `Ice/context` and `Ice/middleware`, the note sits under the "Building the demo" heading instead of before it, because those intros already end with a NOTE block and stacking two blockquotes trips markdownlint MD028.

Running the launchers attaches the application JVM to the terminal, so Ctrl+C reaches the application and the shutdown hooks display their messages (#413).

Drive-by cleanups:

- Standardize on "classpath" (over "class path") in the IceGrid READMEs and deployment descriptors.
- Add the missing `java-library plugin` link definition in the IceGrid/icebox README.
- Fix the `Ice/secure` note that still referred to the `--args="--Ice.Trace.Network"` wrapper.

Verified locally: `installDist` builds and the launcher paths match in `IceBox/greeter`, `IceStorm/weather`, and `Ice/optional` (no project overrides the default application name, so the paths generalize), plus an end-to-end run of `Ice/optional` via the launchers, including the server's "Caught Ctrl+C, shutting down..." message on termination — the exact behavior #413 reported missing. Markdownlint passes with the CI invocation.

Closes #868.
Closes #413.

### DIFF
--- a/java/Ice/config/README.md
+++ b/java/Ice/config/README.md
@@ -20,13 +20,13 @@ bin/ directory.
 
 ## Running the demo
 
-First, start the server application:
+First, start the server application from this demo's directory:
 
 ```shell
 ./server/build/install/server/bin/server
 ```
 
-Then, in a separate terminal, start the client application:
+Then, in a separate terminal, start the client application from this demo's directory:
 
 ```shell
 ./client/build/install/client/bin/client
--- a/java/Ice/config/README.md
+++ b/java/Ice/config/README.md
@@ -2,6 +2,9 @@
 
 This demo shows how to configure client and server applications using Ice configuration files.
 
+> [!NOTE]
+> On Windows, run all the commands below in Git Bash or PowerShell; they don't work in the cmd.exe Command Prompt.
+
 ## Building the demo
 
 The demo has two Gradle projects, **client** and **server**, both using the [application plugin].
@@ -9,21 +12,24 @@ The demo has two Gradle projects, **client** and **server**, both using the [app
 To build the demo, run:
 
 ```shell
-./gradlew build
+./gradlew installDist
 ```
+
+This creates a self-contained distribution under build/install/ for each application, with launcher scripts in its
+bin/ directory.
 
 ## Running the demo
 
 First, start the server application:
 
 ```shell
-./gradlew :server:run --quiet
+./server/build/install/server/bin/server
 ```
 
 Then, in a separate terminal, start the client application:
 
 ```shell
-./gradlew :client:run --quiet
+./client/build/install/client/bin/client
 ```
 
 [Application plugin]: https://docs.gradle.org/current/userguide/application_plugin.html
--- a/java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
+++ b/java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
@@ -2,7 +2,6 @@
 
 package com.example.ice.secure.client;
 
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.security.KeyManagementException;
 import java.security.KeyStore;
@@ -42,8 +41,8 @@ class Client {
     /**
      * Creates and initializes an SSLContext for use with the ssl transport.
      *
-     * <p>The SSLContext is configured with the demo CA certificate loaded from a PKCS12 keystore,
-     * which is used as the trust store to validate server certificates.
+     * <p>The SSLContext is configured with the demo CA certificate loaded from a PKCS12 keystore bundled with the
+     * application, which is used as the trust store to validate server certificates.
      *
      * @return the initialized SSLContext
      * @throws RuntimeException if an error occurs during SSL initialization
@@ -52,14 +51,14 @@ class Client {
         try {
             SSLContext sslContext = SSLContext.getInstance("TLS");
             KeyStore keyStore = KeyStore.getInstance("PKCS12");
-            String keyStorePath = "../../../../certs/ca.p12";
 
             // The password for the PKCS12 keystore, hard-coded for simplicity.
             // In a production environment, use a secure method to store and retrieve this password.
             char[] password = "password".toCharArray();
 
-            try (var input = new FileInputStream(keyStorePath)) {
-                keyStore.load(input, password);
+            // The build bundles ca.p12 (from the top-level certs directory) into the application JAR.
+            try (var caCert = Client.class.getResourceAsStream("/ca.p12")) {
+                keyStore.load(caCert, password);
             }
             TrustManagerFactory trustManagerFactory =
                 TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
--- a/java/IceDiscovery/replication/README.md
+++ b/java/IceDiscovery/replication/README.md
@@ -33,7 +33,7 @@ Then, in a separate terminal, start the client application:
 ./client/build/install/client/bin/client --Ice.Trace.Locator
 ```
 
->[!NOTE]
+> [!NOTE]
 > The `--Ice.Trace.Locator` command-line option is optional: it turns on tracing (logging) for locator resolution and
 > helps you understand the locator logic implemented by the IceDiscovery plug-in.
 
--- a/java/IceDiscovery/replication/README.md
+++ b/java/IceDiscovery/replication/README.md
@@ -3,6 +3,9 @@
 This demo illustrates how to configure the IceDiscovery plug-in. The IceDiscovery plug-in allows a client application
 to discover Ice objects without hardcoding any addressing information.
 
+> [!NOTE]
+> On Windows, run all the commands below in Git Bash or PowerShell; they don't work in the cmd.exe Command Prompt.
+
 ## Building the demo
 
 The demo has two Gradle projects, **client** and **server**, both using the [application plugin].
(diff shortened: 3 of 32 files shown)


pepone (commented):
I found two working-directory regressions that should be fixed before merging:

1. `java/Ice/config/README.md:26,32`

   Both launchers are now run from the demo root, but the applications load `server.conf` and `client.conf` relative to their working directories. Consequently, both commands exit immediately with `FileException: failed to open '<name>.conf'`. The previous Gradle `run` tasks used the respective subproject directories.

2. `java/Ice/secure/README.md:26,32`

   These applications load certificates using paths such as `../../../../certs/server.p12`, which are likewise relative to the `server` or `client` subproject directory. From the demo root, both launchers fail with `FileNotFoundException`.

For both demos, the README could change into the appropriate subproject before invoking each launcher, for example:

```shell
cd server
./build/install/server/bin/server
```

Alternatively, the applications could locate or package their runtime files independently of the current working directory.

I reproduced all four failures after successful `installDist` builds. I also scanned the remaining Java demos and found no other application launchers with this kind of runtime working-directory dependency.

bernardnormier:
Thanks for catching these — both fixed in 1ae00183c, without adding `cd` steps to the READMEs:

- **Ice/config**: `server.conf` and `client.conf` moved to the demo root, where the launchers run from, with `IceSSL.DefaultDir` adjusted to `../../../certs`. This matches the C++ port's layout exactly.
- **Ice/secure**: the keystores are now bundled into the application JARs — each subproject's build packages `server.p12` / `ca.p12` from the top-level `certs` directory as a resource, and the code loads them with `getResourceAsStream` instead of `FileInputStream`. The launchers are now independent of the working directory (this diverges slightly from the C++/C# ports, which read the certs via relative file paths, but it's the idiomatic approach for Java JARs).

Both demos re-tested after the fix: full SSL handshake and greeting, with the secure client run from a different directory than the demo root to confirm the working-directory independence.

java/IceDiscovery/replication/README.md
```diff
@@ -3,28 +3,34 @@
...
 ```
 
 Then, in a separate terminal, start the client application:
 
 ```shell
-./gradlew :client:run --quiet --args="--Ice.Trace.Locator"
+./client/build/install/client/bin/client --Ice.Trace.Locator
 ```
 
 >[!NOTE]
```
InsertCreativityHere: Pre-existing, but there's a space missing here:
```suggestion
> [!NOTE]
```

java/IceGrid/locatorDiscovery/README.md
```diff
@@ -6,6 +6,9 @@ LocatorDiscovery plug-in.
 This demo provides a client application that works with the IceGrid/greeter demo and reuses its server components and
 IceGrid configuration.
 
+> [!NOTE]
+> On Windows, run all the commands below in Git Bash or PowerShell; they don't work in the cmd.exe Command Prompt.
+
```
InsertCreativityHere, line 12: Pretty minor, but this README directs users to:
> run `icegridregistry`, `icegridnode`, and configure IceGrid using `icegridadmin`

All of which run fine in `cmd`. So this should use the alternate wording we use in other demos which have a mix of cmd compatible commands:

```suggestion
> [!NOTE]
> On Windows, run all the commands below in Git Bash or PowerShell; not all of them work in the cmd.exe Command
> Prompt.

```

java/Ice/config/README.md
InsertCreativityHere, line 1: In other demos where commands are sensative to `cwd`, we tack on "from this demo's directory:" in the README.
We should do the same here for the `config` demo, since the loading of `.conf` files depends on `cwd`.

For an example of what I'm talking about, see:
- java/Glacier2/greeter/README.md:42
- java/Glacier2/session/README.md:47
- java/IceGrid/greeter/README.md:48,54
- java/IceStorm/weather/README.md:39
- etc.........

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
```diff
@@ -52,14 +51,14 @@ private static SSLContext createSSLContext() {
...
-            String keyStorePath = "../../../../certs/ca.p12";
 
             // The password for the PKCS12 keystore, hard-coded for simplicity.
             // In a production environment, use a secure method to store and retrieve this password.
             char[] password = "password".toCharArray();
 
-            try (var input = new FileInputStream(keyStorePath)) {
-                keyStore.load(input, password);
+            // The build bundles ca.p12 (from the top-level certs directory) into the application JAR.
+            try (var caCert = Client.class.getResourceAsStream("/ca.p12")) {
```
InsertCreativityHere, line 60: Maybe this is getting a little too into the weeds, but if `getResourceAsStream` fails for any number of reasons, it doesn't throw, it just returns `null`, which will be silently accepted.

And we could easily fix this silent-failure condition with:
```suggestion
            try (var caCert = requireNonNull(Client.class.getResourceAsStream("/ca.p12"))) {
```
With an accompanying `import`.

InsertCreativityHere (approved):
Looks good to me!
Just some minor nits.

pepone (approved):

java/Ice/secure/client/src/main/java/com/example/ice/secure/client/Client.java
bernardnormier: The stream can only be null if the JAR was built without the keystore, and the build itself packages it — so this can't happen with the demo as shipped. And if it somehow did, `keyStore.load` throws an NPE right away, so the failure isn't silent. I'd rather keep the demo focused on the Ice code than add error handling for a build-misconfiguration path, so I'll leave it as is.

InsertCreativityHere: Fine with me, but for the record, this is factually false. It silently accepts null with no exception:

> And if it somehow did, keyStore.load throws an NPE right away, so the failure isn't silent.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/KeyStore.html#load(java.io.InputStream,char%5B%5D)
```
Parameters:
    stream - the input stream from which the keystore is loaded, or null
```

bernardnormier: You're right, my mistake — `KeyStore.load` is documented to accept a null stream and simply initializes an empty keystore, so there's no NPE. The actual behavior would be worse than I claimed: the client would build a truststore that trusts nothing and fail later with an `SSLHandshakeException` ("unable to find valid certification path"), and the server would end up with no key material and fail the handshake on its end. Thanks for the correction.

PR #874, opened 2026-08-10
Fix the runtime DLL copy on CMake < 3.26

externl:
The C++ demos copy runtime DLLs on Windows with `cmake -E copy -t <dir> <files...>`, but the `-t` argument was only added in CMake 3.26. Since these projects declare a CMake 3.21 minimum, configuring with CMake 3.21 through 3.25 produces a post-build copy step that treats `-t` as a source file and fails.

This replaces every occurrence with the destination-last form (`cmake -E copy <files...> <dir>`), which works with all supported CMake versions. Found while reviewing zeroc-ice/ice#6430, where the same snippet in the documentation was corrected.


pepone (approved):
Verified against CMake 3.25.2: the `-t` form fails, the destination-last form succeeds.

bernardnormier (approved):

Issue #844, opened 2026-08-06
Add "What to look at" sections to demo READMEs


bernardnormier:
Split out from #791 (theme A2) — editorial follow-ups from the internal demo-review report.

Add a short "What to look at" section to each demo README pointing at the key code line(s) that demonstrate the featured concept.

Called out in the review (across the languages where applicable): greeter, config, optional, bidir, cancellation, glacier2 session, glacier2 callback, IceDiscovery, IceGrid query, invocation_timeout, IceStorm weather, and the DataStorm demos (key/sample filters, partial updates, custom encoding).

The guiding principle from the review: make each demo's featured Ice concept easy to identify without over-explaining unrelated production concerns.

bernardnormier:
Closing as not planned: we prefer to keep the demo READMEs short and focused on build/run instructions. The demos are small and heavily commented, so readers are better served by going straight to the code than by an extra routing section in each README.

PR #876, opened 2026-08-11
Highlight SSL configuration in the Config demo READMEs

### LINKED ISSUE #846: Highlight SSL configuration in the Config demo READMEs
Split out from #791 (theme A4) — editorial follow-ups from the internal demo-review report.

The Config demos undersell their SSL content. Add a "What to look at" pointing to the SSL / certificate / keychain / `Ice.Default.Protocol` / tracing properties in the config files (and note that client and server must agree on the protocol): `python`, `csharp`, `swift`, `java`, `matlab`, `php`, `ruby` Config demos.

bernardnormier:
Fixes #846.

The Config demo READMEs never mentioned that these demos run over SSL. Rather than adding a "What to look at" section (see #844), this extends the intro sentence to say what the demonstrated configuration actually does: select `ssl` as the default transport protocol, specify the certificates used by IceSSL, and turn on tracing.

The tracing clause is omitted for the php and ruby demos, whose config files only show the tracing properties as comments.

### DIFF
--- a/cpp/Ice/config/README.md
+++ b/cpp/Ice/config/README.md
@@ -1,7 +1,7 @@
 # Config
 
-This demo shows how to configure your client and server applications using Ice configuration files. The configuration
-we demonstrate sets up SSL: the properties in `client.conf` and `server.conf` select `ssl` as the default transport
+This demo shows how to configure client and server applications using Ice configuration files. The configuration we
+demonstrate sets up SSL: the properties in `client.conf` and `server.conf` select `ssl` as the default transport
 protocol and specify the certificates used by IceSSL; `server.conf` also turns on tracing.
 
 ## Ice prerequisites
--- a/cpp/Ice/config/README.md
+++ b/cpp/Ice/config/README.md
@@ -1,6 +1,8 @@
 # Config
 
-This demo shows how to configure your client and server applications using Ice configuration files.
+This demo shows how to configure your client and server applications using Ice configuration files. The configuration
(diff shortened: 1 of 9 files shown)


pepone (approved):
Two optional notes, neither blocking:

- In the four client+server demos where the client's `IceSSL.Trace.Security` is commented out (cpp, csharp, java, swift), tracing is only turned on in `server.conf`. If you want per-file precision: "…and, in `server.conf`, turn on tracing." Author's call.
- Spotted while verifying: `python/Ice/config/client/client.conf` ends with an uncommented `Ice.Trace.Protocol=1` — no explanatory comment, and no other Config demo sets it (leftover from #367). Worth removing in a follow-up.

cpp/Ice/config/README.md
```diff
@@ -1,6 +1,8 @@
 # Config
 
-This demo shows how to configure your client and server applications using Ice configuration files.
+This demo shows how to configure your client and server applications using Ice configuration files. The configuration
+we demonstrate sets up SSL: the properties in `client.conf` and `server.conf` select `ssl` as the default transport
+protocol and specify the certificates used by IceSSL; `server.conf` also turns on tracing.
```
InsertCreativityHere: This README says "... configure your client and server ..."
But all the others say "... configure client and server ..."

I don't have anything against `your` here. But we should use the same wording all the READMEs, not have a slightly different wording in the C++ one.

InsertCreativityHere (approved):
Looks good to me, just one minor inconsistency.

Issue #870, opened 2026-08-07
Ice/secure: use a dedicated keychain with correct key permissions instead of the login keychain


pepone:
`cpp/Ice/secure/README.md` tells macOS users to import the server identity into the **default (login) keychain**:

```shell
security import ../../../certs/server.p12 -P password
```

That is the configuration Ice moved away from. Per zeroc-ice/ice#5343:

> This change is required to negotiate forward-secret (ECDHE) cipher suites on the server side: a private key in the
> login keychain cannot complete an ECDHE handshake, so with the previous default a macOS server could not use
> forward secrecy.

So this demo is now walking users into the exact setup that cannot do forward secrecy on the server side. That
matters more than usual here, because Secure is the demo people read to learn how to set SSL up properly.

The demo can't inherit the fix from zeroc-ice/ice#5343 — that applies when Ice imports `IceSSL.CertFile` itself,
whereas this demo does its own manual import and its own identity lookup
(`Util::loadSecIdentityWithMatchingSubject`, `cpp/Ice/secure/UtilSecureTransport.cpp`). It has to be fixed here.

## Proposal

Show how to create a **dedicated keychain with the right key permissions** and use that instead of the login
keychain. This is the pattern ZeroC's own IceSSL test suite uses (`scripts/tests/IceSSL/configuration.py`):

```shell
security create-keychain -p password demo.keychain
security set-keychain-settings demo.keychain          # no auto-lock while the demo runs
security unlock-keychain -p password demo.keychain
security import ../../../certs/server.p12 -f pkcs12 -A -P password -k demo.keychain
```

`-A` grants key access to any application without prompting. That is the "correct permissions" part, and it is also
what lets the server actually use the key. Scoping it to the demo binary with `-T ./build/server` instead of `-A` is
the tighter option if we would rather not model a blanket ACL.

Cleanup becomes a single line the README can offer, instead of asking people to hunt through Keychain Access:

```shell
security delete-keychain demo.keychain
```

This keeps the demo's existing scope decision intact — the import is still a manual `security` command, and the
[Keychain Services] API is still out of scope (README lines 103-104).

## This needs a code change too

`Util::loadSecIdentityWithMatchingSubject` builds its query without `kSecMatchSearchList`, so it searches the user's
**default search list** — which a dedicated keychain is not on. Verified: after creating `demo.keychain` and
importing `server.p12` into it, the demo's query verbatim returns `errSecItemNotFound` (-25300); adding
[...]

PR #878, opened 2026-08-12
Stop pinning IceSSL.Keychain in the macOS Config demos

### LINKED ISSUE #869: Stop pinning IceSSL.Keychain in the macOS Config demos
The macOS Config demos pin a keychain file next to the demo:

```ini
IceSSL.Keychain=server.keychain
IceSSL.KeychainPassword=password
```

That keychain persists between runs, and macOS keys certificate records on (type, issuer DN, serial number). The
test certificates are regenerated with `certs/makecerts.sh`, which resets the serial file to `1000` on every run
(tracked upstream as zeroc-ice/ice#6488), so each new leaf reuses the serial of the one it replaces (client
`0x1000`, server `0x1001`) under an unchanged issuer DN. A refreshed certificate is therefore rejected as a
duplicate by any keychain that still holds its predecessor, and the demo keeps using the old — now expired —
certificate.

Reproduced with a throwaway keychain, using the certificates before and after #865:

```shell
security import old/server.p12 -k demo.keychain -P password   # "1 identity imported. 1 certificate imported."
security import new/server.p12 -k demo.keychain -P password   # "2 certificates imported."  <- no identity added
security find-identity demo.keychain
#  1) A822CF4C... "ca.server" (CSSMERR_TP_CERT_EXPIRED)       <- the new leaf never landed
```

A direct `SecItemAdd` of the new leaf into that keychain returns `errSecDuplicateItem` (-25299). The new CA
certificate, whose serial *did* change, adds normally. The `*.keychain` files are gitignored, so they survive a
`git pull` and there is nothing to hint at the cause.

## Proposal

Stop setting `IceSSL.Keychain` / `IceSSL.KeychainPassword` in the Config demos. Since zeroc-ice/ice@f46dd2890d
("Use a temporary keychain for the macOS SSL transport certificate", zeroc-ice/ice#5343), leaving `IceSSL.Keychain`
unset makes the macOS transport import `IceSSL.CertFile` into a temporary keychain that is deleted when the
communicator is destroyed. Every run then starts from a fresh import, which removes this failure mode entirely and
drops two properties from the demo configuration.

Files to change:

- `cpp/Ice/config/client.conf`, `cpp/Ice/config/server.conf`
- `swift/Ice/Config/client.conf`, `swift/Ice/Config/server.conf`
- `python/Ice/config/client/client.conf`, `python/Ice/config/server/server.conf`
- `php/Ice/config/client.conf`
- `ruby/Ice/config/client.conf`

The now-dead `*.keychain` entries in the matching `.gitignore` files can go too.

## Prerequisite

This change depends on the Ice-side behavior from zeroc-ice/ice#5343, which was cherry-picked to the Ice `3.8`
branch (zeroc-ice/ice@31f1112f8f) and shipped in Ice 3.8.2. On Ice 3.8.0 and 3.8.1, an unset `IceSSL.Keychain`
falls back to `SecKeychainCopyDefault()`, i.e. the user's **login** keychain: it pollutes the user's keychain with
the demo certificates, hits the same duplicate-serial problem in a place users are less likely to clean up, and —
per that commit's own comment — a private key in the login keychain cannot complete a forward-secret (ECDHE)
handshake on the server side.

So the change can land on `main` and be cherry-picked to the ice-demos `3.8` branch right away. The accepted
consequence is that, once backported, the macOS Config demos on the ice-demos `3.8` branch require Ice >= 3.8.2
rather than any 3.8.x.

## Related

- #865 — refreshes the expired certificates; this issue is the follow-up that keeps the *next* refresh from being
  masked by a stale keychain.
- #846 — highlight SSL configuration in the Config demo READMEs; whoever touches these `.conf` files should keep the
  two in sync.
- #870 — the equivalent problem in `cpp/Ice/secure`, which manages its own keychain and so is not fixed by this
  change. It needs a dedicated keychain rather than the login keychain, for forward-secrecy reasons as well.

bernardnormier:
Fixes #869.

On macOS, the Config demos pinned a keychain file next to the demo (`IceSSL.Keychain` / `IceSSL.KeychainPassword`). That keychain persists between runs, and because `certs/makecerts.sh` reuses leaf serial numbers (zeroc-ice/ice#6488), a refreshed certificate is rejected as a duplicate by a keychain that still holds its predecessor, leaving the demo on the old, expired certificate.

Since Ice 3.8.2 (zeroc-ice/ice#5343), leaving `IceSSL.Keychain` unset imports `IceSSL.CertFile` into a temporary keychain deleted when the communicator is destroyed, so every run starts from a fresh import.

Changes:

- Drop `IceSSL.Keychain` and `IceSSL.KeychainPassword` from the 8 Config demo `.conf` files (cpp, swift, python client+server; php, ruby client). The remaining `IceSSL.Password` line gets a comment matching the file style.
- Remove the now-dead "macOS keychain files" block (`*.keychain`, `*.fl*`) from all 25 `.gitignore` files that carried it — the 5 Config demos plus 20 Swift demo directories that never set `IceSSL.Keychain`.

The keychain references in `cpp/Ice/secure` are intentionally untouched; that demo manages its own keychain and is tracked as #870.

Tested on macOS with the Ice 3.9 nightly: built `cpp/Ice/config` and ran server + client end to end over SSL — handshake succeeds and no keychain files are created next to the demo.


pepone (approved):
🤖 **Claude:** Verified the temporary-keychain behavior against the Ice `3.8` sources (`SecureTransportEngine.cpp`: unset `IceSSL.Keychain` + `CertFile` → keychain under `_CS_DARWIN_USER_TEMP_DIR`, deleted in the engine destructor) and swept the branch: all 8 `.conf` setters and all 25 `.gitignore` blocks on `main` are covered, and nothing else references the removed files or properties.

One suggestion: `csharp/Ice/Config/{Client,Server}` and `matlab/Ice/config` leave `IceSSL.Password` uncommented — consider adding the same comment there so the Config family stays consistent.

The two red CI jobs (`rewrite`, `ubuntu-24.04`) are the same transient Gradle-distribution download EOF in Java demos this PR doesn't touch; a re-run should clear them.

PR #879, opened 2026-08-12
Fix low-severity Python demo audit findings

### LINKED ISSUE #816: [Audit-Low] Python demos: UTC sensor timestamps, customError prints args tuple, misc config/naming nits
> **AI-generated findings** — batch of Low-severity Python items from the [verification report][verif] (NEW-L06, NEW-L07, NEW-L16) accompanying the [2026-08-04 ice-demos correctness audit][audit], by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`). Code is pinned to `main` at [`99afbe3`][commit]. **Not yet human-triaged** — please verify before acting.

## 1. Weather sensor timestamps in UTC; every other port uses local time (NEW-L06)

https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/python/IceStorm/weather/sensor/main.py#L52-L54

C++ formats with `localtime`, C# uses `DateTime.Now`, Java `LocalTime.now()`, etc. — a mixed-language station display shows timestamps disagreeing by the UTC offset.

## 2. customError prints the exception-args tuple (NEW-L07)

https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/python/Ice/customError/client/main.py#L34-L37

`{exception.args}` renders the raw tuple — `message = '('Invalid credentials…',)'` with tuple punctuation — where the C#/Ruby/PHP siblings print the bare string. `Ice.DispatchException` exposes no `message` attribute (verified against Ice for Python's `LocalExceptions.py`); use `{exception}` / `str(exception)`. Line 40 handles the Slice-defined `GreeterException` correctly.

## 3. Misc (NEW-L16 b–c)

- [`python/Ice/greeter/server_amd/pyproject.toml#L2`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/python/Ice/greeter/server_amd/pyproject.toml#L1-L3) names the AMD *server* project `"client"`.
- [`python/Ice/config/client/client.conf#L40`](https://github.com/zeroc-ice/ice-demos/blob/99afbe3075144465da7ac53f6caabb98d13a4db7/python/Ice/config/client/client.conf#L38-L41) carries a stray `Ice.Trace.Protocol=1` found in no other conf file in the repo; the config demo dumps protocol traces its README never mentions.

## Note (unverified lead, §7 of the verification report)

`python/Ice/invocation_timeout/server` does not cancel the outstanding 60-second dispatch on Ctrl+C (the Swift counterpart calls `cancelOutstandingTasks()` for exactly this reason); whether Ctrl+C actually hangs depends on communicator-destroy semantics and was not exercised — worth a quick manual run while touching these files.

bernardnormier:
Fixes #816.

- **Weather sensor timestamps**: `python/IceStorm/weather/sensor` now formats timestamps with local time (`datetime.now()`), matching the C++, C#, and Java sensors.
- **customError client**: print `{exception}` instead of `{exception.args}` for `Ice.DispatchException`, so the message is the bare string (as in the C#/Ruby/PHP clients) rather than a tuple.
- **greeter server_amd**: rename the pyproject project from `client` to `ice-greeter-server-amd`, following the naming convention of the other Python demos.
- **invocation_timeout server**: cancel the outstanding slow-greeter dispatch on Ctrl+C, mirroring the Swift server. Without this, communicator destruction waits for the 60-second dispatch to complete — reproduced with SIGINT: the server printed "Caught Ctrl+C, shutting down..." and then hung for the remaining ~57 seconds. With the fix it exits in ~2 seconds, and the canceled dispatch throws `Ice.UnknownException` like its Swift counterpart.

The stray `Ice.Trace.Protocol=1` in `python/Ice/config/client/client.conf` (item 3 of the issue) is already gone on `main`.

Verified by running the customError and invocation_timeout demos end to end; `ruff check`, `ruff format --check`, and `pyright` pass on all touched demos.


pepone:
Nice cleanup overall — the weather-sensor local-time switch, the `{exception}` fix, and the pyproject rename all look correct, and the invocation_timeout hang is real and worth fixing. The comments below are all about the new cancellation code; the first one means a badly-timed Ctrl+C can still reproduce the 60-second hang this PR sets out to fix.

### Cancellation logic (`invocation_timeout`)

**1. Latch the cancelled state — the current cancel is one-shot** — `server/chatbot.py:43`
Ice queues each `greet` coroutine on the loop via `run_coroutine_threadsafe` from a thread-pool thread, so a request that arrives just before Ctrl+C can be queued but not yet registered in `_outstandingTasks` when `cancelOutstandingTasks()` runs. The cancel loop misses it, the coroutine then starts a fresh 60-second sleep during communicator destruction, and the server hangs anyway. (Reordering with `communicator.shutdown()` doesn't close this — deactivation can't retract an already-queued dispatch, and `destroyAsync()` initiates shutdown regardless.) The C++/C#/Java siblings avoid this because their cancellation primitive stays signalled (`shared_future`/`CancellationToken`/`CompletableFuture`). The Python equivalent: set a `self._canceled` flag in `cancelOutstandingTasks()` and check it at the top of `greet`, raising the `UnknownException` immediately — everything runs on the event-loop thread, so no locking is needed.

**2. Don't swallow external cancellation** — `server/chatbot.py:36`
`except asyncio.CancelledError` also catches the case where the *dispatch task itself* is cancelled from outside — awaiting a task delegates outer cancellation into it, and the inner task then reports `cancelled() == True` too, so the inner task's state can't distinguish the two. Converting that case to `Ice.UnknownException` breaks asyncio's re-raise-on-cancel contract; e.g. a second Ctrl+C while a dispatch is pending has `asyncio.run()`'s shutdown cancel the dispatch tasks and produces "unhandled exception during asyncio.run() shutdown" noise. Re-raise when the dispatch task has a pending cancellation request (`asyncio.current_task().cancelling()`) before converting.

**3. The bundled client doesn't handle the new reply** — `client/main.py:36`
The client only catches `Ice.InvocationTimeoutException`, but the cancellation path now replies with `Ice.UnknownException` — a reply the old always-sleeps-60s server could never deliver inside the 4-second window. Ctrl+C on the server while the slow call is in flight kills the client with a traceback. The C#/Java/Swift demos catch `UnknownException` explicitly ("Caught UnknownException, as expected") in a deliberate server-cancellation invocation.

**4. Consider gating the task tracking on the delay** — `server/chatbot.py:31`
The tracking machinery runs for every dispatch, including the regular `Chatbot(0.0)` that serves all normal requests, but only the slow greeter's set is ever cancelled — and if that first Chatbot ever gains a real delay, its tasks are never cancelled on shutdown. Swift gates task creation on `if let delay` (C++/Java gate the wait similarly); `delay: float | None = None` with the bookkeeping skipped when unset would match and keep the fast path readable.

### Other

**5. `replyStatus` still prints a raw int** — `customError/client/main.py:36`
The `{exception}` fix normalizes the message, but Python prints the numeric reply status where the C# client prints the enum name (its `replyStatus` is a `ReplyStatus` enum). Converting via `Ice.ReplyStatus` with a fallback to the raw int (custom reply statuses outside the enum are legal) would match the C# output for this demo's statuses.

**6. README / teaching flow** — `invocation_timeout/README.md`
The server now demonstrates dispatch cancellation (new "greet dispatch canceled" output on Ctrl+C), but nothing surfaces it: the README still describes only invocation timeouts, and unlike the Swift demo — whose client makes a deliberate indefinite invocation, tells the user to interrupt the server, and catches the resulting `UnknownException` — the Python client only hits this path by accident. Worth either documenting the behavior or mirroring the Swift client flow (which would also naturally resolve #3).

### Nits (chatbot.py / main.py)

- `asyncio.ensure_future` → `asyncio.create_task` (the rest of the python/ tree uses `create_task`).
- The cancel message f-string is duplicated between the `print` and the `raise`; also consider `raise Ice.UnknownException(...) from None` to avoid chaining the `CancelledError` into logged tracebacks.
- `set[asyncio.Task[None]]` is more precise; `cancelOutstandingTasks` is missing `-> None`.
- The comment above `main.py:22` reads "a regular greater" → "a regular greeter".

bernardnormier:
Thanks for the thorough review! All points addressed in the latest commit — the client now follows the Swift demo flow (deliberate indefinite invocation with Ctrl+C on the server), and I verified the shutdown behavior end to end, including a double Ctrl+C.

pepone (approved):
All points addressed — thanks for the client rework, it reads exactly like the Swift demo now. Verified end to end: Ctrl+C during the indefinite `carol` invocation cancels both outstanding dispatches, the server exits immediately, and the client prints `Caught UnknownException, as expected`; a double Ctrl+C shuts down cleanly as well.

